PR #97, opened 2014-09-18
Unit tests less verbose and with preinstallation to avoid tests to core addons

dreispt:


bwrsandman:
The changes to test server make this patch tough to rebase. I will close this with hopes that the issue is no longer present. If it is, the patch probably needs to be rethought due to the changes done to the script.

PR #99, opened 2014-09-19
Fix pylint error on server_environment

### LINKED ISSUE #98: False negative on Traceback
Current https://github.com/OCA/server-tools shows green for server tests, but you can see in the log a traceback `ValueError: External ID not found in the system: base.module_mi`

Seen in this log: https://travis-ci.org/dreispt/server-tools/jobs/35732685

dreispt:
Fixes #98.


bwrsandman:
This is already implemented in test_server.py

Issue #112, opened 2014-10-18
[REF] Proposal to check "print" setence and "pdb" sentence


moylop260:
Many times our developers add "print" or "pdb" sentence to traceback a code.
However, it is susceptible to forget to remove when doing commit and push
With ast is easy check this sentence.
See example here:
https://github.com/Vauxoo/pylint-conf/blob/master/check_ast/check_print_and_pdb.py

I can make a PR to add "custom_lint" script.

How do you think?

PR #113, opened 2014-10-20
[IMP][pylint-conf] Add deprecated modules: pdb, pudb. And enable fails W0402 to check it #112

moylop260:
More info here:
https://github.com/OCA/maintainer-quality-tools/issues/112

We need enable 
W0402(deprecated-module) to detect conf
+[IMPORTS]
+deprecated-modules=pdb,pudb
But show fail only if module is installed.

And we need enable 
F0401(import-error)
If you try use deprecated modules without install depends.

And now "import openerp" show a import-error then I add odoo directory to PYTHONPATH to detect this import.

### DIFF
--- a/travis/cfg/travis_run_pylint.cfg
+++ b/travis/cfg/travis_run_pylint.cfg
@@ -6,7 +6,7 @@ cache-size=500
 
 [MESSAGES CONTROL]
 disable=all
-enable=E0101,E1124,E1306,E1601,I0013,W0101,W0102,W0104,W0105,W0109,W0403,W0404,W1111,W1401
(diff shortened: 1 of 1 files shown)


bwrsandman:
:+1:

bwrsandman:
Waiting for an example test before merge

yvaucher:
:+1:

oscarolar:
[Build Status](https://travis-ci.org/vauxoo-dev/server-tools) Here is the build status of server-tools using MQT/113 directly from branch.

bwrsandman:
Good work!

PR #152, opened 2014-12-02
[ADD] docker file and shippable compatibility

moylop260:
Docker odoo image
Building here
[ocaimages/oca-odoo](https://registry.hub.docker.com/u/ocaimages/oca-odoo/builds_history/91750/)


docker/odoo/Dockerfile
```diff
@@ -0,0 +1,38 @@
...
+
+MAINTAINER "Vauxoo"
+
+#  Fix log odoo warning "unable to set lang fr_FR"
+RUN locale-gen fr_FR \
+    && locale-gen en_US.UTF-8 \
+    && dpkg-reconfigure locales \
+    && update-locale LANG=en_US.UTF-8 \
+    && update-locale LC_ALL=en_US.UTF-8 \
+    && echo 'LANG="en_US.UTF-8"' > /etc/default/locale
```
ruiztulio: @yvaucher this is a nice solution  when testing o using a single session, but if you use docker-enter to interact with a docker container some env vars are not available for the session so is better for us to define it system wide, that's why we use this way instead of defining an ENV

bwrsandman:
Needs 
- [ ] to enable OCA on shippable
- [ ] A badge in the sample_files/README.md
- [ ] Rebase and conflict resolution for install nightly, no sudo
- [ ] Explanation in the main README.md for what shippable is and how it is used.
Other than that :+1:

dreispt:
@bwrsandman Why enable OCA on Shippable? I believe that this adds that option for projects using MQT, but OCA can keep using TravisCI only.

I tested Shippable on some OCA projects, but on the free plan builds are not parallel, so they ended up taking too long compared to TravisCI.
On the other hand TravisCI only supports GitHub, so it's not an option for projects hosted on BitBucket, that's way IMO the Shippable option is relevant.

bwrsandman:
This PR is about adding shippable as a test option. It therefore needs an OCA org on shippable.

As for the merits of Shippable, I cannot say. I didn't test it.

@moylop260 can you chime in about the runtime?

moylop260:
Feel free to continue this work

blaggacao:
@moylop260 Is there a way to extract the part needed for https://github.com/vauxoo/travis2docker ?

I ask in the context of this: https://github.com/OCA/maintainer-tools/issues/146

I want to be able to craft a simple tooling for newcomers to start with docker development and make it available here: https://github.com/OCA/maintainer-tools/wiki/OCA-Contributor-Toolbox-(OCA-CT)

PR #158, opened 2014-12-18
Update sample files

dreispt:
- Unit tests test dependencies, so no there's no point in running twice.
- Make 8.0 the default in the sample files.


yvaucher:
@dreispt I think I had to put VERSION="8.0" with LINT_TEST="1"
As in .travis.yml install step you have command using VERSION

e.g. git clone https://github.com/OCA/stock-logistics-workflow ${HOME}/stock-logistics-workflow -b ${VERSION}

Not sure how to get rid of it as when we want to check LINT, we don't need to clone other repo.

yvaucher:
@dreispt Ok if we consider adding https://github.com/OCA/maintainer-quality-tools/pull/159 my last comment is obselete.

:+1:

dreispt:
@yvaucher I think your comment is relevant. I'll propose the addition of the version to the LINT line.

PR #159, opened 2015-01-05
cross repository dependency management

### LINKED ISSUE #105: add a dependencies.txt file at repository level
I'm working on the configuration of the runbot for OCA project, and this requires setting up the dependencies of the various projects, a process which is quite tedious, and also related to the writing of `.travis.yml` where such dependencies need to be manually coded. 

I was thinking of a metadata file at the project level describing the inter-repository dependencies which could be used to help in the process. In the simplest form, it would be a plain text file, with 1 repository per line giving the names of the repository on which the current project depends. The scripts launched by travis could be updated so that in case the file is present, they would checkout the listed repository with the  `${VERSION}` branch.  And having this information available would allow for an easier configuration of the runbot (at the moment, I'm grepping for `git clone` lines in `.travis.yml`...)

If people agree with this idea, I'll propose an implementation.

gurneyalex:
use repo/oca_dependencies.txt rather than an explicit list of checkouts in
.travis.yml

closes #105

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,11 +12,9 @@ To setup the TravisCI continuous integration for your project, just copy the
 content of the [`/sample_files`](https://github.com/OCA/maintainer-quality-tools/tree/master/sample_files)
 to your project’s root directory.
 
-If your project depends on other OCA/Github repositories simply add the following under `before_install` section:
+If your project depends on other OCA/Github repositories, create a file called `oca_dependencies.txt` at the root of your project and list the dependencies in there, one per line:
 
-    install:
-      - git clone https://github.com/OCA/a_project_x ${HOME}/a_project_x -b ${VERSION}
-      - git clone https://github.com/OCA/a_project_y ${HOME}/a_project_y -b ${VERSION}
+    project_name optional_repository_url optional_branch_name
 
 The addons path used will automatically consider these repositories.
 
--- a/sample_files/oca_dependencies.txt
+++ b/sample_files/oca_dependencies.txt
@@ -1,0 +1,15 @@
+# List the OCA project dependencies, one per line
+# Add a repository url and branch if you need a forked version
+#
+# Examples
+# ========
+#
+# To depend on the standard version of sale-workflow, use:
+# sale-workflow
+#
+# To explicitely give the URL of a fork, and still use the version specified in
+# .travis.yml, use:
+# sale-workflow https://github.com/OCA/sale-workflow
+#
+# To provide both the URL and a branch, use:
+# sale-workflow https://github.com/OCA/sale-workflow branchname
--- a/tests/test_repo/oca_dependencies.txt
+++ b/tests/test_repo/oca_dependencies.txt
@@ -1,0 +1,5 @@
+# List the OCA project dependencies, one per line
+# Add a repository url and branch if you need a forked version
+
+partner-contact https://github.com/OCA/partner-contact 8.0
+
--- a/travis/clone_oca_dependencies
+++ b/travis/clone_oca_dependencies
@@ -56,7 +56,7 @@ def parse_depfile(depfile):
 def git_checkout(home, reponame, url, branch):
     checkout_dir = osp.join(home, reponame)
     if not osp.isdir(checkout_dir):
-        command = ['git', 'clone', url, '-b', branch, checkout_dir]
+        command = ['git', 'clone', '-q', url, '-b', branch, checkout_dir]
         _logger.info('Calling %s', ' '.join(command))
         subprocess.check_call(command)
     return checkout_dir
--- a/travis/clone_oca_dependencies
+++ b/travis/clone_oca_dependencies
@@ -1,0 +1,100 @@
+#!/usr/bin/python
+"""Usage: clone_oca_dependencies [<checkout_dir> <build_dir>]
+
+Arguments:
+
+checkout_dir: the directory in which the dependency repositories will be cloned
+build_dir: the directory in which the tested repositories have been cloned
+
+If no arguments are provided, default to the layout used in the OCA travis
+configuration.
+
+The program will process the file oca_dependencies.txt at the root of the
+tested repository, and clone the dependency repositories in checkout_dir,
+before recursively processing the oca_dependencies.txt files of the
+dependencies.
+
+The expected format for oca_dependencies.txt:
+
+* comment lines start with # and are ignored
+* a dependency line contains:
+  - the name of the OCA project
+  - (optional) the URL to the git repository (defaulting to the OCA repository)
+  - (optional) the name of the branch to use (defaulting to ${VERSION})
+"""
+from __future__ import print_function
+import sys
+import os
+import os.path as osp
+import subprocess
+import logging
+
+
+_logger = logging.getLogger()
+
+
+def parse_depfile(depfile):
+    deps = []
+    for line in depfile:
+        line = line.strip()
+        if not line or line.startswith('#'):
+            continue
+        parts = line.split()
+        repo = parts[0]
+        if len(parts) > 2:
+            branch = parts[2]
+        else:
+            branch = os.environ.get('VERSION', '8.0')
+        if len(parts) > 1:
+            url = parts[1]
+        else:
+            url = 'https://github.com/OCA/%s' % repo
+        deps.append((repo, url, branch))
+    return deps
+
+
+def git_checkout(home, reponame, url, branch):
+    checkout_dir = osp.join(home, reponame)
+    if not osp.isdir(checkout_dir):
+        command = ['git', 'clone', url, '-b', branch, checkout_dir]
+        _logger.info('Calling %s', ' '.join(command))
+        subprocess.check_call(command)
+    return checkout_dir
+
+
+def run(home, build_dir):
+    dependencies = []
+    processed = set()
+    for repo in os.listdir(build_dir):
+        _logger.info('examining %s', repo)
+        processed.add(repo)
+        depfilename = osp.join(build_dir, repo, 'oca_dependencies.txt')
+        dependencies.append(depfilename)
+    for depfilename in dependencies:
+        try:
+            with open(depfilename) as depfile:
+                deps = parse_depfile(depfile)
+        except IOError:
+            deps = []
+        for depname, url, branch in deps:
+            _logger.info('* processing %s',  depname)
+            if depname in processed:
+                continue
+            processed.add(depname)
+            checkout_dir = git_checkout(home, depname, url, branch)
+            new_dep_filename = osp.join(checkout_dir, 'oca_dependencies.txt')
+            if new_dep_filename not in dependencies:
+                dependencies.append(new_dep_filename)
+
+
+if __name__ == '__main__':
+    if len(sys.argv) == 1:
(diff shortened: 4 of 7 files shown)


gurneyalex:
See https://github.com/OCA/vertical-ngo/pull/107 for a test of this implementation.

travis/clone_oca_dependencies
```diff
@@ -0,0 +1,100 @@
...
+            deps = []
+        for depname, url, branch in deps:
+            _logger.info('* processing %s',  depname)
+            if depname in processed:
+                continue
+            processed.add(depname)
+            checkout_dir = git_checkout(home, depname, url, branch)
+            new_dep_filename = osp.join(checkout_dir, 'oca_dependencies.txt')
+            if new_dep_filename not in dependencies:
+                dependencies.append(new_dep_filename)
```
guewen: I don't like to modify a list when we are iterating through it with a `for` loop.
I prefer to use something like:
```
    while dependencies:
        depfilename = dependencies.pop()
        ...
        if new_dep_filename not in dependencies:
            dependencies.append(new_dep_filename)
```
(not exactly the same result as `dependencies.pop()` will process the dependencies from last to first but I don't think it is an issue here)

gurneyalex: Appending to a list in a for loop is ok. Removing is ok too, as long as you remove *after* the current iteration element (which is also why appending to a list is perfectly safe: in this case I'm using the list as a queue of elements to be processed).

guewen: Ok :raised_hands:

guewen:
:+1:

travis/clone_oca_dependencies
```diff
@@ -0,0 +1,100 @@
...
+        repo = parts[0]
+        if len(parts) > 2:
+            branch = parts[2]
+        else:
+            branch = os.environ.get('VERSION', '8.0')
+        if len(parts) > 1:
+            url = parts[1]
+        else:
+            url = 'https://github.com/OCA/%s' % repo
+        deps.append((repo, url, branch))
```
nbessi: Just a matter of taste but I prefer named tuple for this kind of usage

sample_files/oca_dependencies.txt
```diff
@@ -0,0 +1,15 @@
...
+#
+# To depend on the standard version of sale-workflow, use:
+# sale-workflow
+#
+# To explicitely give the URL of a fork, and still use the version specified in
+# .travis.yml, use:
+# sale-workflow https://github.com/OCA/sale-workflow
+#
+# To provide both the URL and a branch, use:
+# sale-workflow https://github.com/OCA/sale-workflow branchname
```
nbessi: Don't we want to also have hash support ?

gurneyalex: I don't think so: I have not seen any `.travis.yml` files in the OCA specifying a hash for a dependency branch. I expect most of the time to have only the branch name in the file.

yvaucher:
@gurneyalex in case of LINT check only do we want to clone deps ? Could we have a way of avoiding this unnecessary step ?

gurneyalex:
@yvaucher this is already taken care of in `travis_install_nightly` (https://github.com/OCA/maintainer-quality-tools/blob/master/travis/travis_install_nightly#L6) and I have not changed this part.

yvaucher:
@gurneyalex oh right thanks

Then :+1:

nbessi:
:+1:

README.md
```diff
@@ -12,11 +12,9 @@ To setup the TravisCI continuous integration for your project, just copy the
 content of the [`/sample_files`](https://github.com/OCA/maintainer-quality-tools/tree/master/sample_files)
 to your project’s root directory.
 
-If your project depends on other OCA/Github repositories simply add the following under `before_install` section:
+If your project depends on other OCA/Github repositories, create a file called `oca_dependencies.txt` at the root of your project and list the dependencies in there, one per line:
 
-    install:
-      - git clone https://github.com/OCA/a_project_x ${HOME}/a_project_x -b ${VERSION}
-      - git clone https://github.com/OCA/a_project_y ${HOME}/a_project_y -b ${VERSION}
+    project_name optional_repository_url optional_branch_name
```
pedrobaeza: You should put `OCA_project_name` and put the other two parameters enclosed by brackets, that it's the common terminology for regular expressions with optional parts.

sample_files/oca_dependencies.txt
pedrobaeza: Same comments as in doc.

tests/test_repo/oca_dependencies.txt
```diff
@@ -0,0 +1,5 @@
+# List the OCA project dependencies, one per line
+# Add a repository url and branch if you need a forked version
+
+partner-contact https://github.com/OCA/partner-contact 8.0
```
pedrobaeza: This should be also `partner-contact` to show a simple example. Then, you can put another line with a complete example for a personal repository.

travis/clone_oca_dependencies
pedrobaeza: I don't think we should add dependencies recursively for each dependency, because main project can depend on a module of a repository that has dependencies because of another modules, but not the one we need.

Making this, we can increase a lot the build time (specially with those branches that has a lot of commits or branches - this will get worse with each Odoo version).

Each repository maintainer should resolve manually the dependency chain and add it to main project.

lepistone: I am with @pedrobaeza here. Repos are loose groups of modules, and I am not very keen to work too much on repo dependencies (a concept I'd like to get rid of eventually). We already have the proprietary odoo dependency resolution, and we're adding even another one.

Still, I understand how that makes life a bit easier, and I am not blocking it. Thanks!

tests/test_repo/oca_dependencies.txt
gurneyalex: this is a test, not an example. The example is in the sample_files/oca_dependencies.txt

README.md
gurneyalex: that's the common way for USAGE strings output when `--help` is passed as a command line option... In the context of an example configuration I tend to find my version clearer. 

Waiting for more feedback on this.

travis/clone_oca_dependencies
gurneyalex: I disagree. 

This will reduce breakage when other repositories' external dependencies change. And I personnally hate having to manually track these when a computer can perfectly do the job in my stead (and should do so).

@pedrobaeza Built time increase argument is void. The resulting clones are currently exactly the same as the one we are manually configuring, and the checkout time of OCA repositories is negligible compared to creating the base database.

pedrobaeza: @gurneyalex, no if added dependencies (for new modules) are not used by our repository. For example, spanish localization uses base_location of partner-contact repository. Only that module. But module _partner_relations_ needs web repository because it uses _web_m2x_options_ and _web_tree_many2one_clickable_, and thus this dependency is added at partner-contact level.

The way you make it, it will download also project web, and the corresponding dependencies for it, and so on recursively, needing to download practically all OCA repos for each Travis test, when I only need for spanish localization to download partner-contact repo.

dreispt: I also agree with Pedro. I would rather see "nested" dependencies be explicitly declared. I also hope this gives incentive keep dependencies to a minimum and try to stay away from a "monolith".

nbessi:
@dreispt @pedrobaeza can we merge this one ?

pedrobaeza:
My concern about dependency resolution hasn't been addressed as far as I can see.

travis/clone_oca_dependencies
```diff
@@ -0,0 +1,100 @@
...
+    return deps
+
+
+def git_checkout(home, reponame, url, branch):
+    checkout_dir = osp.join(home, reponame)
+    if not osp.isdir(checkout_dir):
+        command = ['git', 'clone', '-q', url, '-b', branch, checkout_dir]
```
guewen: I think we can use `depth=1` to clone faster.

guewen:
> My concern about dependency resolution hasn't been addressed as far as I can see.

Is it only a matter of build time? I'm not sure that it is really a concern...
```
git clone git@github.com:OCA/sale-workflow.git --depth=1
```
takes 2 seconds on my computer

pedrobaeza:
--depth=1 solution seems good to me

yvaucher:
Ok I had a look at our repo and their dependencies, and I changed my mind.

I don't think there is a case we would download a monolithic source base.

This for few reasons:

- Our dependency tree has a low depth
- Topics alwost always implies that if you are pulling a repo you will need all its dependencies.
- Many repo still have no dependencies

Thus for the few case we would pull extra repo, it's not worth the time to have to write down everywhere the exact list needed.

We can later improve this automatically searching deps in `__openerp__.py` certainly be writing down somewhere the hierarchy to avoid too many connexion to github repo.

In 9.0 we won't have this issue anymore if 1 repo = 1 module

StefanRijnhart:
@yvaucher that is a big __if__, isn't it?

yvaucher:
@StefanRijnhart For 9.0 more I think about it and more I'm convinced it would be the best way to go to manage dependencies.

dreispt:
Saying that a repo depends on another is a misleading way to put it.
It's more accurate to say that some modules in a repos depend on another.

We can have a module in repo A depend on a module in repo B, and another module in repo B depend on a module in repo C.
This makes repo A "depend" on repo B but not on repo "C". 
In math jargon, it's a non transitive relation. :alien: 

Repos are organized around common functional areas.
Inter-repo dependencies are to be avoided, or at least, not incentivized.
We should work to keep complexity down having modules as decoupled as possible,
Inter-repo dependencies are still to be expected, but mostly in border areas.

I believe that having a border module in A depends on a border module of B that needs a module in C should be a rare situation.
And since it's not desirable to have too much coupling, I would prefer to have these cases explicitly identified.

So I don't sympathize with the auto discovery of dependencies.
But feel free to go on if I'm the only one with this opinion.

pedrobaeza:
@dreispt, I think similar to you, but this can be minimized if the auto-discovery doesn't penalize performance.

lepistone:
So a few of us, me included, are a bit skeptical of bringing the idea of repo dependencies too far.

Recently, I have been working on modules like `stock-dropshipping-dual-invoice` that clearly deal with stock, sales and purchases, so it is easy to see what **modules** it depends on, but the **repos** it touches (and makes dependent on each other) are much less clear. With modules like that, the idea of independent repositories seems to me difficult, useless and counterproductive.

It is indeed useful to group modules for human reasons (determining the maintainer, listing them clearly), but technically, I'd love to get rid of module dependencies one day (with one-module repos, probably).

TL;DR here what @gurneyalex proposes addresses a problem we have today, and he did the work already, so I am :+1: for merging anyway! Is anyone against?

yvaucher:
needs a rebase

bwrsandman:
:+1: I am for as well

It can help greatly for external tools such as travis, but also runbot, documentation, translation and all sorts of other test/utility servers.

dreispt:
@gurneyalex I think you have 4 thumbs up.

max3903:
:+1:

PR #171, opened 2015-03-27
set sudo: required to avoid container based builds

### LINKED ISSUE #170: red build because travis_install_nightly did not install the dependencies
https://travis-ci.org/OCA/stock-logistics-transport/jobs/56092821 is red because lxml is not available. 

In the logs I notice:

    $ travis_install_nightly
    Installing Odoo https://github.com/odoo/odoo/archive/8.0.tar.gz
    2015-03-27 13:30:42 URL:https://codeload.github.com/odoo/odoo/tar.gz/8.0 [79515718/79515718] -> "odoo.tar.gz" [1]
    Installing Python dependencies
    sudo: must be setuid root
    sudo: must be setuid root
    You are using pip version 6.0.7, however version 6.0.8 is available.
    You should consider upgrading via the 'pip install --upgrade pip' command.
    Getting dependencies

It looks like the version of pip is causing the install not to take place.

gurneyalex:
see http://docs.travis-ci.com/user/workers/container-based-infrastructure/
closes #170

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 # WARNING:
 # Do not use this file as a template for you repository TravisCI configuration
 # Instead use the files under the "/sample_files" directory for that purpose.
-
+sudo: required
 language: python
 
 python:
--- a/sample_files/.travis.yml
+++ b/sample_files/.travis.yml
@@ -1,3 +1,4 @@
+sudo: required
 language: python
 
 python:


guewen:
:+1:

pedrobaeza:
Shouldn't you put `sudo: false`?

pedrobaeza:
I thought it would be better to use container based builds.

gurneyalex:
@pedrobaeza I'll help you write the script to change the .travis.yml files once you've figured out how to get green container based builds. Right nown what is needed is to stop the false Red builds we get because this is completely blocking our workflow. 

I just made a mass commit on all OCA projects (branches 6.1, 7.0 and 8.0) to update the travis configuration.

lepistone:
@pedrobaeza @gurneyalex: I agree with you both:

- today we need to make our builds work as before, fast
- tomorrow I imagine we'll get faster builds with containers, using a wheel cache for instant pip installs (so no need to sudo apt-get install lxml). The wheel cache works on my machine.

dreispt:
:+1:
Merge?

pedrobaeza:
For now it has to be this way, so I merge.

PR #180, opened 2015-04-30
Run coveralls conditionally

guewen:
Currently, the `coveralls` command runs on every build: the tests builds but the lint one as well.

Yesterday, I added a new build on OCA/connector#60 which builds the documentation; I added environment variables to switch off `coveralls` (https://github.com/OCA/connector/commit/bc55f9aab88927a75624778fd02922ab426264e6#diff-354f30a63fb0907d4ad57269548329e3R15).

After a few exchanges on Twitter with @bwrsandman and @pedrobaeza, I engaged myself to propose a change in the MQT to allow `coveralls` to run conditionnally. So here it is.

This PR is backward compatible. To be effective on the projects, we'll need to replace `coveralls` by `travis_after_tests_success` in all the `.travis.yml` files. If we agree on this change, I can make a script that creates pull requests for that (and in any case, until the line is changed, the builds will still continue to work as of today).

`coveralls` is now activated when the environment variable `TESTS` is `"1"`. As it is a new variable, it is set by default to `"1"` to be backward compatible. However, it allows to switch off the tests on a case by case basis. Example for the  build of the docs: OCA/connector#61

I also proposed an little improvement in the sample file by putting the version in `env: global:`, it don't need to be repeated on every lines (the result is exactly the same).

### DIFF
--- a/sample_files/.travis.yml
+++ b/sample_files/.travis.yml
@@ -8,7 +8,7 @@ env:
   global:
   - VERSION="8.0"
   matrix:
-  - LINT_CHECK="1"
+  - LINT_CHECK="1" TESTS="0"
   - TESTS="1" ODOO_REPO="odoo/odoo" LINT_CHECK="0"
   - TESTS="1" ODOO_REPO="OCA/OCB" LINT_CHECK="0"
 # either use the two lines above or the two below. Don't change the default if
--- a/sample_files/.travis.yml
+++ b/sample_files/.travis.yml
@@ -5,9 +5,12 @@ python:
   - "2.7"
 
 env:
-  - VERSION="8.0" LINT_CHECK="1"
-  - VERSION="8.0" ODOO_REPO="odoo/odoo" LINT_CHECK="0"
-  - VERSION="8.0" ODOO_REPO="OCA/OCB" LINT_CHECK="0"
+  global:
+  - VERSION="8.0"
+  matrix:
+  - LINT_CHECK="1"
+  - TESTS="1" ODOO_REPO="odoo/odoo" LINT_CHECK="0"
+  - TESTS="1" ODOO_REPO="OCA/OCB" LINT_CHECK="0"
 # either use the two lines above or the two below. Don't change the default if
 # it's not necessary (it is only necessary if modules in your repository can't
 # be installed in the same database. And you get a huge speed penalty in your
@@ -27,4 +30,4 @@ script:
   - travis_run_tests
 
 after_success:
-  coveralls
+  - travis_after_tests_success
--- a/sample_files/.travis.yml
+++ b/sample_files/.travis.yml
@@ -6,17 +6,17 @@ python:
 
 env:
   global:
-  - VERSION="8.0"
+  - VERSION="8.0" TESTS="0" LINT_CHECK="0"
   matrix:
-  - LINT_CHECK="1" TESTS="0"
-  - TESTS="1" ODOO_REPO="odoo/odoo" LINT_CHECK="0"
-  - TESTS="1" ODOO_REPO="OCA/OCB" LINT_CHECK="0"
+  - LINT_CHECK="1"
+  - TESTS="1" ODOO_REPO="odoo/odoo"
+  - TESTS="1" ODOO_REPO="OCA/OCB"
 # either use the two lines above or the two below. Don't change the default if
 # it's not necessary (it is only necessary if modules in your repository can't
 # be installed in the same database. And you get a huge speed penalty in your
 # tests)
-#  - VERSION="8.0" ODOO_REPO="odoo/odoo" UNIT_TEST="1" LINT_CHECK="0"
-#  - VERSION="8.0" ODOO_REPO="OCA/OCB" UNIT_TEST="1" LINT_CHECK="0"
+#  - TESTS="1.0" ODOO_REPO="odoo/odoo" UNIT_TEST="1"
+#  - TESTS="1.0" ODOO_REPO="OCA/OCB" UNIT_TEST="1"
 
 virtualenv:
   system_site_packages: true
--- a/travis/travis_after_tests_success
+++ b/travis/travis_after_tests_success
@@ -5,6 +5,6 @@ set -v
 # if the env variable TESTS is not set, default to 1 to keep backward compatibility
 : ${TESTS:="1"}
 
-if [ "${TESTS}" = "1" ] ; then
+if [ "${TESTS}" = "1" ] && [ "${LINT_CHECK}" != "1" ] ; then
   coveralls
 fi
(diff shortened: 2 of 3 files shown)


sample_files/.travis.yml
```diff
@@ -27,4 +30,4 @@ script:
   - travis_run_tests
 
 after_success:
-  coveralls
```
pedrobaeza: Can't we call it the script `coveralls`, that calls the real coveralls command, instead of renaming all Travis?

sample_files/.travis.yml
```diff
@@ -5,9 +5,12 @@ python:
...
 
 env:
-  - VERSION="8.0" LINT_CHECK="1"
-  - VERSION="8.0" ODOO_REPO="odoo/odoo" LINT_CHECK="0"
-  - VERSION="8.0" ODOO_REPO="OCA/OCB" LINT_CHECK="0"
+  global:
+  - VERSION="8.0"
+  matrix:
+  - LINT_CHECK="1" TESTS="0"
+  - TESTS="1" ODOO_REPO="odoo/odoo" LINT_CHECK="0"
```
yvaucher: I wonder if variable like LINT_CHECK and TEST could go in global setted to "0" to be overwritten

sample_files/.travis.yml
guewen: As we add MQT in the PATH, we could easily shadow `coveralls` indeed. And then, call directly`/usr/local/bin/coveralls` (path to check on travis). 

But it is safe to do so? Can we assume that the `coveralls` path won't change in travis builds? Also, an inconvenience is that it would add a layer of noise and technical debt.

What are the other opinions here?

sample_files/.travis.yml
guewen: Excellent idea. I just checked and it works. Sample file updated.

sample_files/.travis.yml
```diff
@@ -5,15 +5,18 @@ python:
...
+  - LINT_CHECK="1"
+  - TESTS="1" ODOO_REPO="odoo/odoo"
+  - TESTS="1" ODOO_REPO="OCA/OCB"
 # either use the two lines above or the two below. Don't change the default if
 # it's not necessary (it is only necessary if modules in your repository can't
 # be installed in the same database. And you get a huge speed penalty in your
 # tests)
-#  - VERSION="8.0" ODOO_REPO="odoo/odoo" UNIT_TEST="1" LINT_CHECK="0"
-#  - VERSION="8.0" ODOO_REPO="OCA/OCB" UNIT_TEST="1" LINT_CHECK="0"
+#  - VERSION="8.0" ODOO_REPO="odoo/odoo" UNIT_TEST="1"
```
yvaucher: You might want to replace VERSION="8.0" by TESTS="1" from those commented example lines

sample_files/.travis.yml
dreispt: I agree with adding an after success script. It gives us a nice extension point for future needs.

yvaucher:
:+1: I also agree we better not shadow coveralls

We will anyway need to rewrite .travis.yml to add TESTS="1" so we will have to change coverall call as well.

pedrobaeza:
Shadowing the name is not needed to change .travis.yml, because TESTS="1" is the default.

pedrobaeza:
We still can have extension points calling the script `coveralls` instead of `travis_after_tests_success`

sample_files/.travis.yml
bwrsandman: I also agree

travis/travis_after_tests_success
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+
+set -v
+
+# if the env variable TESTS is not set, default to 1 to keep backward compatibility
+: ${TESTS:="1"}
+
+if [ "${TESTS}" = "1" ] && [ "${LINT_CHECK}" != "1" ] ; then
+  coveralls
+fi
```
bwrsandman: In python this could be:

```python
#!/usr/bin/env python

import os
from coveralls import cli as coveralls_cli

if os.environ.get("TESTS") and not os.environ.get("LINT_CHECK"):
    coveralls_cli.main(argv=None)
```

bwrsandman:
The change is good, and with alone, I approve. :+1: 
Though, it would be much better to call the python command using python, wouldn't you agree?

travis/travis_after_tests_success
guewen: Nice, I didn't thought about it. Changed to a Python script.

guewen:
> Though, it would be much better to call the python command using python, wouldn't you agree?

Done!

pedrobaeza:
I insist in reconsider call the script coveralls to avoid changing all .travis.yml, and now that the script is Python, I think it's even better to avoid conflicts.

bwrsandman:
You're not wrong, except that aliasing the script as coveralls is by definition creating a conflict.

We could go either way, I think they would both be equally good/bad solutions.

dreispt:
@pedrobaeza Explicit is better than implicit.

travis/travis_after_tests_success
```diff
@@ -0,0 +1,7 @@
+#!/usr/bin/env python
+
+import os
+from coveralls import cli as coveralls_cli
+
+if os.environ.get('TESTS', True) and not os.environ.get('LINT_CHECK'):
```
dreispt: I think some confs have LINT_CHECK="0", so I suggest using `os.environ.get('LINT_CHECK', False)`

travis/travis_run_tests
```diff
@@ -37,6 +38,9 @@ if __name__ == '__main__':
     if os.environ.get("LINT_CHECK", "1") == "1":
         tests.append(['test_flake8'])
         tests.append(['test_pylint'])
-    if os.environ.get("LINT_CHECK", None) != "1":
+    # Execute tests by default. Disable them on lint checks
+    if (os.environ.get("LINT_CHECK") != "1" and
```
dreispt: `os.environ.get('LINT_CHECK', '0')` ?

guewen: if LINT_CHECK is is not set, `.get()` returns `None` which is different than `"1"`, so it is not needed to default to `'0'`.

travis/travis_after_tests_success
guewen: You are right, we have "0" and "1" strings, and `TESTS` could be `0` too, so we have to check the values.

travis/travis_run_tests
dreispt: OK

dreispt:
:+1:

lmignon:
:+1:

dreispt:
@pedrobaeza Can we merge?

pedrobaeza:
OK, if all of you prefer not shadowing 'coveralls' name and modify Travis, go ahead.

PR #182, opened 2015-05-04
Automatic discovery and installation of module dependencies

dreispt:
This PR embeds into MQT a stable version of the [`ooenv`](https://github.com/dreispt/ooenv/blob/master/README.md) tools, to mimic `virtualenv` + `pip install` for Odoo modules.

`ooenv` adds thow CLI commands to Odoo:

- `env` creates a directory to hold an Odoo installation and it's modules, *a la* virtualenv.
- `get` adds modules to a created environment, downloading their Git repo if necessary. For that an index mapping module names to Git repo URLS is needed. That's the purpose for the index files making up most of the line diff. They were automatically generated by a very simple script.

This PR will be easier to read by viewing each commit separately.


travis/travis_install_nightly
```diff
@@ -15,22 +15,39 @@ fi
...
 cp ${HOME}/maintainer-quality-tools/cfg/.coveragerc .
 
-echo "Getting dependencies"
-clone_oca_dependencies
+echo "==== Getting dependencies ===="
+cd ${TRAVIS_BUILD_DIR}  # Maybe redundant, but just in case go into the build dir
+if [ "${VERSION}" -lt "8.0" ] ; then
+    clone_oca_dependencies
+else
+    ${ODOO_PATH}/odoo.py env . --index=${MQT_PATH}/ooenv/oca-index
```
dreispt: This creates an env in the current directory, using as index the `${MQT_PATH}/ooenv/oca-index` directory.

travis/travis_install_nightly
```diff
@@ -15,22 +15,39 @@ fi
...
+    echo "Possible modules:"
+    find . -maxdepth 1 -mindepth 1 -type d -printf '%f\n'
+    find . -maxdepth 1 -mindepth 1 -type d -printf '%f\n' | xargs ./odoo.sh get --verbose
```
dreispt: This runs the `get` command on every module in the repo to test. Of course the modules are already there, but their dependencies will still be processes, so this will trigger the download and availability of foreign modules we may depend on.

bwrsandman:
Is this still relevant when we have oca_dependencies.txt? I like that approach since it is a sort of self documenting file on each repo for the dependencies you might need.

I am not sure how this method would be maintainable...

dreispt:
This eliminates the additional manual dependency maintenance that oca_depencies.txt implies (except for central index refresh when new OCA modules are merged).

Not implemented, but if needed a more exact repo dependency list (as in oca_dependencies.txt), could easily be calculated from a selected module list.

pedrobaeza:
I think keeping this kind of index is very painful and error-prone. There are already too much places to maintain when adding modules/repos.

PR #184, opened 2015-05-05
Easier to read log outputs

dreispt:


dreispt:
[Example](https://travis-ci.org/OCA/maintainer-quality-tools/jobs/61280404) - compare with the [former](https://travis-ci.org/OCA/maintainer-quality-tools/jobs/61277401) output.

travis/test_flake8
```diff
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-set -v
+# set -v  # verbose
```
bwrsandman: You can remove this entirely

dreispt: I decided to comment out instead, to make it easier to re-enable when developing new features.

bwrsandman: The reason for that line, when it was originally introduced was so when devs sees a test failing, they can see the flake command being run.
At the time it was beneficial since the parameters told you which rules were disabled.

Now we use config files, so the rules don't show up in the test output and those devs need to dig around to see the config files.

It would be better to have our code style guidelines explicitly say which rules should be respected before a pull request with commands (or hooks) that they can run on their own machine.

Now, if there is a real use for it other than what I described, then we can keep it. Do you use that bash directive when developing new features?

dreispt: OK, no problem: I'll remove it.

bwrsandman:
:+1:

yvaucher:
:+1:

PR #186, opened 2015-05-08
Export translation files to Transifex

Ehtaga:
This PR allows to export all the PO files of the OCA repositories to Transifex, thanks to Travis.

The export is only done for official commits and not pull requests. It's not executed with default tests. You need to add a line in your `.travis.yml` env matrix that set the `TRANSIFEX` variable to `"1"`.
For example:

```
env:
  matrix:
    - LINT_CHECK="1"
    - TRANSIFEX="1"
    - TESTS="1" ODOO_REPO="odoo/odoo"
    - TESTS="1" ODOO_REPO="OCA/OCB"
```

The script files are in Python, they don't use subprocess calls. We use the [txclib](https://pypi.python.org/pypi/transifex-client) library and the export translation files functions of Odoo. The `test_server` script has been renamed in `test_server.py` in order to import some of its functions useful for the Transifex export. 

~~It assumes that the translation project already exists on Transifex with the following slug : `<repo_name>-<version>` (with "-" instead of "." in the version string). Transifex OCA organization should have one project created by repo and by version.
Unfortunately, the project creation part can't be done remotely.~~

A 'transbot' account has been created with the rights to publish on Transifex OCA projects.
Its password needs to be encrypted and added to the `.travis.yml` file of each project. It is generated thanks to the project name, thus it is a different one in each project. To get it, you need to run the command:
```
$ travis encrypt TRANSIFEX_PASSWORD=your-password -r owner/project
```

If the script runs on new modules, it creates the corresponding resources on Transifex (a resource <=> a module). If not, it updates them. In order to do that, we need to use the 0.11b3 version of tx (0.10 doesn't create new resources).

We use the following transifex commands to publish our files:

* `tx init` -> Initialize the connection settings with the Transifex server. It creates a new `.tx` directory in travis local directory with a `config` file that will hold the mapping between local files and resources in Transifex. At this step, it only contains the host url.
A `.transifexrc` file is also created in travis home directory for storing user login and password.

* `tx set` -> This command allows to set the association between the local files and the Transifex server resources. We use `--auto-local` option which looks at the project's folders locally and sets the `config` file to reflect that structure. We set the files type (PO), the resource slug, the source file and the source language. We also give an expression that will reflect the structure of all translation files.
For instance, the source file would be `/my/path/my-module/i18n/my-module.pot` and the expression would be `/my/path/my-module/i18n/<lang>.po`

* `tx push` -> This command allows to upload source and translated files in order to create/update resources on the Transifex project.

The `travis_transifex` script has got the following structure:
```
  -> setup server (same function than in test_server script) and install the modules
  -> init Transifex project
  -> for each module in repo
    -> download .pot source file
    -> set the mapping between local files and Transifex resources
  -> push source and translation files to Transifex
```

In order to be able to download .pot source file for a given module, we need to connect to Odoo. This is done in `OdooXContext` classes. We implemented an abstract class (`_OdooBaseContext`) with `__init__`, `__enter__` and `__exit__` functions. It allows us to temporarily add Odoo server path to system path, add addon paths to config and init all the necessary Odoo objects. It also contains a function that exports source translation contents for a given addon. Each version has a corresponding `OdooXContext` class.

The script has been tested for 8.0 and 7.0 versions. If you want it to run for former versions, you're free to take a look at it. :) You will need to create a subclass of _OdooBaseContext.

Concretely, for each repo, you need to:
- Create a project for each version in Transifex, if not already done. At this step, you can choose which language will be handled by Transifex.
- Encrypt the transifex 'transbot' password with the owner/repo
- Add the sample `.travis.yml` file to your repo. If you want Travis to perform Transifex exports, it is necessary to have the TRANSIFEX_USER variable, the encrypted TRANSIFEX_PASSWORD and a line in the matrix with TRANSIFEX="1"
- Commit and enjoy

Here are some screenshots of the translation files export of OCA/l10n-canada, with a test account:

transifex0

transifex1

transifex2

transifex3

An interesting future work would be to add a script that gets all the translations done on Transifex and automatically pushes them to Github.


bwrsandman:
Code review and test: :+1:

max3903:
:+1: 

No test

bwrsandman:
A little info about encrypting keys:

The tool can be gotten from ruby gem:
```
gem install travis
```
 http://docs.travis-ci.com/user/encryption-keys/

bwrsandman:
I've run into an issue with the transifex project names when testing on l10n-canada:

The transifex project created by @max3903 is of the form: `l10n-canada-80`, `l10n-canada-70`. The travis test assumes $(REPONAME)-$(BRANCHNAME).replace(".", "-") and tries to push to `l10n-canada-8-0`, `l10n-canada-7-0`. Those projects belong to TestOrganization2 which I assume is @Ehtaga's test account.

@pedrobaeza has l10n-spain-7-0, I assume and it would work, but other than that, OCA has no other projects.

I am worried about future naming conflicts, perhaps we should agree to name our transifex projects under the following nomenclature:
OCA-$(REPONAME)-$(BRANCHNAME).replace(".", "-")

bwrsandman:
@Ehtaga could you look at https://travis-ci.org/OCA/l10n-canada/jobs/61826119, that test should have failed but gives a false positive.

travis/travis_transifex.py
```diff
@@ -0,0 +1,121 @@
...
+    print()
+    print(yellow('Initializing Transifex project'))
+    init_args = ['--host=https://www.transifex.com',
+                 '--user=%s' % transifex_user,
+                 '--pass=%s' % transifex_password]
+    commands.cmd_init(init_args, path_to_tx=None)
+    path_to_tx = utils.find_dot_tx()
+
+    repo_name = "%s-%s" % (travis_repo_slug.split("/")[1],
+                           odoo_version.replace('.', '-'))
```
bwrsandman: This is the nomenclature used.

Perhaps, this should be:
```python
    repo_name = "%s-%s" % (
        travis_repo_slug.replace("/", "-"),
        odoo_version.replace('.', '-'),
    )
```

Ehtaga: `travis_repo_slug` contains: `owner_name/repo_name` and we only need `repo_name`, that's why we split it.

bwrsandman: I think we should keep owner_name in the transifex project due to possible name conflicts on transifex like the `web` repo

Ehtaga:
You're right for the project names.
I will add a variable in `.travis.yml` in the case we need to change it. As I already created a project named `l10n-canada-8-0` with my test account, you're not able to create another one with the same name with OCA account. Even if we agree for a nomenclature, this could be problematic in the future if another organization already owns a project with the same name as ours.

About the false positive test:
I have configured the push command to skip errors. Otherwise, this command exits with an error code if:
- we export an empty source file, with no string to translate
- we export a translation file without any translations (the `msgstr` fields are still `""`)
Those are common cases and we don't want the export to stop and fail for that.

travis/travis_transifex.py
Ehtaga: Ok, I'll change the default project name for `owner`-`repo_name`-`version`.

bwrsandman:
Here is how I propose we avoid the naming conflicts for projects:

capture d ecran de 2015-05-11 10-00-26

bwrsandman:
Here's a live result: https://www.transifex.com/projects/p/OCA-l10n-canada-8-0/

sample_files/README.md
```diff
@@ -1,5 +1,6 @@
 [![Build Status](https://travis-ci.org/OCA/${REPO_NAME}.svg?branch=${BRANCH_NAME})](https://travis-ci.org/OCA/${REPO_NAME})
 [![Coverage Status](https://coveralls.io/repos/OCA/${REPO_NAME}/badge.png?branch=${BRANCH_NAME})](https://coveralls.io/r/OCA/${REPO_NAME}?branch=${BRANCH_NAME})
+[![Transifex Status](https://www.transifex.com/projects/p/${REPO_NAME}-${BRANCH_NAME_WITH_DASH}/chart/image_png)](https://www.transifex.com/projects/p/${REPO_NAME}-${BRANCH_NAME_WITH_DASH})
```
bwrsandman: Include ORG_NAME too

bwrsandman: The graph might look better in the description such as:
```markdown

Translation Status
------------------
[![Transifex Status](https://www.transifex.com/projects/p/OCA-l10n-canada-8-0/chart/image_png)](https://www.transifex.com/projects/p/OCA-l10n-canada-8-0)
```
See: https://github.com/OCA/l10n-canada/blob/7.0/README.md

yvaucher:
Great!

LGTM :+1:

sample_files/README.md
yvaucher: Agreed, I was confused using trying the link without putting OCA

.travis.yml
```diff
@@ -11,17 +11,21 @@ virtualenv:
...
+  - TESTS="1" INCLUDE="test_module,second_module" OPTIONS="--log-level=debug"  # OPTIONS usage example
+  - TESTS="1" INCLUDE="test_module,second_module" INSTALL_OPTIONS="--log-level=info"
+  - TESTS="1" INCLUDE="test_module,second_module" UNIT_TEST="1"
+  - TESTS="1" VERSION="7.0" INCLUDE="test_module,second_module" ODOO_REPO="OCA/OCB"  # ODOO_REPO usage example
+  - TESTS="1" VERSION="6.1" INCLUDE="test_module,second_module"
+  - TESTS="1" VERSION="" INCLUDE="test_module,second_module"  # test fallback to script argument
+  # - TESTS="1" VERSION="6.0"  # Test script not ready for v6.0
   - LINT_CHECK="1"
-  - VERSION="7.0" INCLUDE="test_module,second_module" LINT_CHECK="0"
+  - TESTS="1" VERSION="7.0" INCLUDE="test_module,second_module"
```
veloutin: What is the purpose of all these lines? test_module, second_module and broken_module do not exist in here.

To me it seems that this is all stuff for the travis sample file.

Ehtaga: Those modules are in tests/test_repo

travis/test_server.py
```diff
@@ -163,24 +226,26 @@ def main():
...
-                ] + install_options
-    print(" ".join(cmd_odoo))
-    subprocess.check_call(cmd_odoo)
+    setup_server(dbtemplate, odoo_unittest, tested_addons, server_path,
+                 addons_path, install_options)
 
     # Running tests
     database = "openerp_test"
+
+    cmd_odoo_test = ["coverage", "run",
```
veloutin: Are we sure coverage is always available? It isn't in requirements.txt

bwrsandman: It is always with travis python environements

travis/travis_transifex.py
```diff
@@ -0,0 +1,123 @@
...
+        print(yellow_light("WARNING! Transifex user not defined- "
+              "exiting early."))
+        return 1
+
+    if not transifex_password:
+        print(yellow_light("WARNING! Transifex password not recognized- "
+              "exiting early."))
+        return 1
+
+    travis_home = os.environ.get("HOME", "~/")
```
veloutin: os.path.expanduser is never called on the string, using ~/ will likely result in a OSError, no such file or directory.

travis/test_server.py
veloutin: Ah, didn't look for coveralls! Thanks!

veloutin:
Good job! :+1:

Ehtaga:
Add automatic creation of Transifex project if it doesn't already exists. This part uses the Transifex REST API. It allows to check if the variables are correct in the .travis.yml and if the project_slug doesn't already exist on Transifex (owned by another organization for example).
Unfortunately, it's not possible to check if a user or an organization do exist on Transifex, and to check their access rights. So if the authentication fail, you have to check manually if it is due to the user name, the organization name, or the project slug.

Add also security feature when installing requests. This installs following extra packages: pyOpenSSL, ndg-httpsclient and pyasn1. It avoids InsecurePlatformWarning when using slumber.

travis/travis_transifex.py
```diff
@@ -0,0 +1,157 @@
...
+            print('Transifex project has been successfully created.')
+        except exceptions.HttpClientError:
+            print('Transifex organization: %s' % transifex_organization)
+            print('Transifex username: %s' % transifex_user)
+            print('Transifex project slug: %s' % transifex_project_slug)
+            print(red('Error: Authentication failed. Please verify that '
+                      'Transifex organization, user and password are '
+                      'correct. If so, this project slug may be owned by '
+                      'another organization. You can change these variables '
+                      'in your .travis.yml file.'))
```
bwrsandman: Could you print the exception, or even better, reraise it?

pedrobaeza:
@bwrsandman, I haven't had enough time yet to try, but I'll try this weekend. Sorry for the delay.

bwrsandman:
I was able to successfully create a transifex project: https://travis-ci.org/OCA/program/jobs/62722253#L187 https://www.transifex.com/projects/p/OCA-program-7-0/

No setup needed on transifex side

This is the changeset which was needed: https://github.com/OCA/program/compare/95eb7628c0419525af0ca83144b6ccc5cbe8702c...2b339d8fafa994eb9d71e46b816307054b724cc8

Quite simple:
1. Update MQT (or accept this pull request)
2. Add `TRANSIFEX="1"` to the testing matrix and `TRANSIFEX="0"` to other builds
3. Define `TRANSIFEX_USER` and secure `TRANSIFEX_PASSWORD` with `travis encrypt TRANSIFEX_PASSWORD=************* -r OCA/REPONAME` replacing REPONAME

Someone could mass edit the builds to add `TRANSIFEX_USER` and `TRANSIFEX_PASSWORD` to every branch in OCA with no side effects other than having two more environment variables. I can generate the secure strings pretty easily for each project.

The only step needed after, would be to accept the pull request and add the `TRANSIFEX="1"` test

travis/travis_transifex.py
```diff
@@ -0,0 +1,160 @@
...
+    print()
+    print(yellow("Creating Transifex project if it doesn't exist"))
+    auth = (transifex_user, transifex_password)
+    api_url = "https://www.transifex.com/api/2/"
+    api = API(api_url, auth=auth)
+    project_data = {"slug": transifex_project_slug,
+                    "name": transifex_project_name,
+                    "source_language_code": "en",
+                    "description": transifex_project_name,
+                    "repository_url": "http://www.transifex.com",
```
bwrsandman: This isn't right, should be `"https://github.com/%s" % travis_repo_slug`

yvaucher:
@bwrsandman Using global, you could set TRANSIFEX="0" by default and only set it to "1" on a job

Like it was done in last travis template : https://github.com/OCA/maintainer-quality-tools/blob/master/sample_files/.travis.yml

.travis.yml
dreispt: Why do you have to add TESTES="! on all lines? Isn't that the default already?

yvaucher: @dreispt TESTS is set to "0" by default

I find convenient to see which lines are test putting this `TESTS="1"` a beginning of line.

sample_files/.travis.yml
```diff
@@ -11,9 +11,21 @@ addons:
 
 env:
   global:
-  - VERSION="8.0" TESTS="0" LINT_CHECK="0"
+  - VERSION="8.0" TESTS="0" LINT_CHECK="0" TRANSIFEX="0"
+  - TRANSIFEX_USER='transbot@odoo-community.org'
```
dreispt: Not introduced in this PR but: why do we have TESTS="0" on the global env? Doesn't it make more sense to be TESTS="1" and then use TESTS="0" for the build that don't run tests?

.travis.yml
dreispt: IMHO it would make more sense the opposite - to mark with TEST="0" test builds taht don't not running tests. But it's only an opinion.

sample_files/.travis.yml
yvaucher: Thus you would have a noisy `TESTS="0"` on transifex and lint check lines or anything else we would add. We could consider all jobs should be tests unless it isn't. Or that it needs to always be defined.
It's a matter of taste.

guewen: Agree with @yvaucher. If I take as an example the connector, we would have
```
  matrix:
  - LINT_CHECK="1"
  - TESTS="1" ODOO_REPO="odoo/odoo"
  - TESTS="1" ODOO_REPO="OCA/OCB"
  - DOCS="1"
  - TRANSIFEX="1"
  global:
  - secure: ...
  - VERSION="8.0" TESTS="0" LINT_CHECK="0" TRANSIFEX="0" DOCS="0"
```
which seems way cleaner than repeating `TEST="0"` on every lines

.travis.yml
guewen: Agree with @yvaucher. If I take as an example the connector, we would have
```
  matrix:
  - LINT_CHECK="1"
  - TESTS="1" ODOO_REPO="odoo/odoo"
  - TESTS="1" ODOO_REPO="OCA/OCB"
  - DOCS="1"
  - TRANSIFEX="1"
  global:
  - secure: ...
  - VERSION="8.0" TESTS="0" LINT_CHECK="0" TRANSIFEX="0" DOCS="0"
```
which seems way cleaner than repeating `TEST="0"` on every lines

dreispt: @guewen OK, I was writing from the time where most builds were to run tests :wink:

StefanRijnhart:
:+1:, no test. Much appreciated!

sample_files/.travis.yml
```diff
@@ -11,9 +11,21 @@ addons:
...
+  # This line contains the encrypted transifex password
+  # To encrypt transifex password, use
+  # $ travis encrypt TRANSIFEX_PASSWORD=your-password -r owner/project
```
pedrobaeza: Indicate here also how to install: `[sudo] gem install transifex`

pedrobaeza: EDIT: s/transifex/travis

bwrsandman: Which instance of "transifex" needs to be replaced

Since this is already merged, could you make a PR for the gem comment?

pedrobaeza: No, it was my own comment, that I put transifex instead of Travis. OK, I make the PR. I'm configuring Transifex for l10n-spain. I'm doing with my own account. Can you give transbot credentials or I continue with current?

pedrobaeza: Done: https://github.com/OCA/maintainer-quality-tools/pull/193

bwrsandman: I don't have the transbot password on hand, but I will provide you with the encrypted password for l10n-spain tomorrow, when I am at the office.

pedrobaeza: Yeah, that can be helpful. I let the tests then for tomorrow. Good night!

bwrsandman: Here you go #194

jjscarafia:
Hi and thanks for this great tool!
I'm Trying to make it work on this repository https://github.com/ingadhoc/odoo-infrastructure but y could not make it work. Any help is wellcome!

1. I've installed travis (`sudo gem install travis`)
2. I've my transifex user `jjscarafia` with `[transifex_password]`
3. Github 'repository: `https://github.com/ingadhoc/odoo-infrastructure` and github user `jjscarafia`
4. I've run `travis encrypt TRANSIFEX_PASSWORD=[transifex_password] -r ingadhoc/odoo-infrastructure to generate the ecrypted password (much longer than OCA ones, don't know why)
5. I've this [travis file](https://github.com/ingadhoc/odoo-infrastructure/blob/8.0/.travis.yml)
6. I've this transifex organization https://www.transifex.com/ingadhoc/

Then, two cases:
1. If I don't create the project on transifex, I get this error `Error: Authentication failed...` [travis build](https://travis-ci.org/ingadhoc/odoo-infrastructure/jobs/74219174)
2. If I create the project on transifex manually, I do not get an error but no resources are loaded on transifex as you can see [here](https://www.transifex.com/ingadhoc/ingadhoc-odoo-infrastructure-8-0/dashboard/) but I get some errors on [travis build](https://travis-ci.org/ingadhoc/odoo-infrastructure/jobs/74220681) logs)

Any help will be appreciated. Thanks in advance!

bwrsandman:
Don't know if this is related, but I cannot see your organization on transifex

moylop260:
https://www.transifex.com/ingadhoc/
`webpage not found`

jjscarafia:
I thinks that's because yours users are not allowed, I've not found any way to make it Public. 
Samething happens to me with "https://www.transifex.com/oca/" from a user that don't belong to OCA  org.
image

I've add you both to ingadhoc organization, now you should be able to see it

Thanks for the response!

pedrobaeza:
Correct link is https://www.transifex.com/organization/ingadhoc, and please, don't continue dirtying this PR. Open a new issue and continue the talk there.

Ehtaga:
We all have this problem currently, you are not the only one.
This is due to this line:
`Creating instance:`
`/home/travis/odoo-8.0/openerp-server -d openerp_i18n --log-level=warn --stop-after-init --addons-path /home/travis/build/ingadhoc/odoo-infrastructure,/home/travis/odoo-8.0/addons --init base`

The modules aren't installed, only `base` is, so the script isn't able to export any translations. This is due to recent pull request dealing with the preinstalled modules.

Pedro is right, this conversation should continue in a new issue.

jjscarafia:
Thanks you all for your answers! Will try your changes and continue on that or a new issue.
Sorry for the posting in the wrong place!

PR #187, opened 2015-05-11
Remove sudo requirements and use addon apt packages

### LINKED ISSUE #175: travis container based builds
It would be nice to be able to use Travis container based builds for OCA projects. 

To do this we need to get rid of sudo in our build scripts. 

I've noticed https://github.com/travis-ci/apt-package-whitelist on http://docs.travis-ci.com/user/apt/ and filed the following issues:

* [X] https://github.com/travis-ci/travis-ci/issues/3631 
* [X] https://github.com/travis-ci/travis-ci/issues/3630 

to get python-lxml and expect-dev whitelisted. 

once this is done, we can add these lines in .travis.yml for all projects

    addons:
        apt:
            packages:
                - python-lxml
                - expect-dev

and remove the `sudo apt-get install` commands from our build script.

bwrsandman:
Allows use of container based builds which are a lot quicker to load

Closes  #175

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,11 +1,16 @@
 # WARNING:
 # Do not use this file as a template for you repository TravisCI configuration
 # Instead use the files under the "/sample_files" directory for that purpose.
-sudo: required
 language: python
 
 python:
   - "2.7"
+
+addons:
+  apt:
+    packages:
+      - expect-dev  # provides unbuffer utility
+      - python-lxml  # because pip installation is slow
 
 virtualenv:
   system_site_packages: true
--- a/sample_files/.travis.yml
+++ b/sample_files/.travis.yml
@@ -1,8 +1,13 @@
-sudo: required
 language: python
 
 python:
   - "2.7"
+
+addons:
+  apt:
+    packages:
+      - expect-dev  # provides unbuffer utility
+      - python-lxml  # because pip installation is slow
 
 env:
   global:
--- a/travis/travis_install_nightly
+++ b/travis/travis_install_nightly
@@ -21,10 +21,6 @@ echo "Installing Odoo ${ODOO_URL}"
 wget -nv -O odoo.tar.gz ${ODOO_URL}
 tar -xf odoo.tar.gz -C ${HOME}
 
-echo "Installing Python dependencies"
-sudo apt-get -qq -y install expect-dev  # provides unbuffer utility
-sudo apt-get -qq -y install python-lxml  # because pip installation is slooow
-
 # Workaround to force using system site packages (see https://github.com/Shippable/support/issues/241#issuecomment-57947925)
 rm -f $VIRTUAL_ENV/lib/python2.7/no-global-site-packages.txt
 


gurneyalex:
:+1: many thanks for doing the PR. If your tests are successful I'll update the OCA projects.

pedrobaeza:
@gurneyalex, take into account that you cannot mass-update Travis files, because there are some other dependencies installed by hand. For example, https://github.com/OCA/reporting-engine/blob/8.0/.travis.yml#L16

bwrsandman:
@pedrobaeza `xlwt` doesn't seem to be in the [whitelist](https://github.com/travis-ci/apt-package-whitelist/blob/master/ubuntu-precise) perhaps you could open an issue with travis-ci, then it could be added to the addons

pedrobaeza:
Maybe xlswriter, that's is in the whitelist, is the same package, but what I mean is that it can be simply automated for all repos, but it needs to be checked if there is another dependency.

yvaucher:
Automatically creating a PR for each project should resolve your concerns @pedrobaeza

bwrsandman:
If the change is done in a mass-commit kind of way, the pip line won't affected, it will continue to be slow as usual.

However, after that is done, you can remove the pip install and add `xlswriter` to the addons in the .travis.yml

pedrobaeza:
@yvaucher, thanks for the tip. Indeed, that's the solution.

@bwrsandman, doesn't that command need sudo?

bwrsandman:
The addons section is the equivalent to doing `sudo apt-get install xlswriter` but at the container level. Sudo is not required.

See: https://github.com/OCA/maintainer-quality-tools/pull/187/files#diff-3db2c5a2752be648ec9dd9fa0627be47R10

Maybe I should add a helpful comment?

bwrsandman:
If you meant: does pip install require sudo, the answer is no, otherwise it would have been `sudo pip install`

I assume the travis environment has an alias of the type `alias pip='pip --user'`

pedrobaeza:
Thanks, it was the second question what I mean. So at first sight, it will work. OK, let's see the result of this PR!

bwrsandman:
:smile:

pedrobaeza:
:+1: Thanks

.travis.yml
```diff
@@ -1,12 +1,17 @@
...
-sudo: required
 language: python
 
 python:
   - "2.7"
 
+addons:
+  apt:
+    packages:
+      - expect-dev
```
dreispt: Preserve the comment: ` # provides unbuffer utility`

dreispt: I would be great if we could ditch `unbuffer`. I don't recall exactly why using something like `tee` instead didn't cut it...

bwrsandman: I could be wrong but probably because of the colour log? Odoo doesn't have a flag to force coloured log and tee will make it think it's not in a tty, therefore it will not use colours, which makes debugging a failing test a pain.

dreispt: Thanks Sandy, that rings a bell - it was to preserve the coloured log.

bwrsandman: Should I add that to the comment?

nbessi:
If I'm correct this will not disable the access to the sudo command and if a sudo is found in travis in fallback to classical infrastructure. Some repo may have a mandatory need of sudo so we do not want do remove this possibility.  

If it behave as expected it is :+1: for me

.travis.yml
dreispt: Well, can't do any harm...

bwrsandman:
A mandatory need for sudo would be rare, the use of `addons` and sudo-less pip would be strongly recommended.

Otherwise, you can enable sudo directly in the project with `-sudo: required`, but as far as the travis install nightly is concerned there is no need for sudo.

I think we have 3 approves here, are we ready to merge?

bwrsandman:
I am getting `ImportError: No module named lxml` on projects with `sudo required`

https://travis-ci.org/OCA/management-system/jobs/62440427#L145

PR #191, opened 2015-05-21
Add python libs to container directives

bwrsandman:
This should speed up the container based builds by taking time out of the pip install command and using ready-made containers which add virtually no time.

These packages were not taken out of travis/requirements.txt for backwards compatibility. The speedup should still be there since in those cases, pip will see the dependency installed and not attempt to install them.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,6 +11,8 @@ addons:
     packages:
       - expect-dev  # provides unbuffer utility
       - python-lxml  # because pip installation is slow
+      - python-simplejson
+      - python-serial
 
 virtualenv:
   system_site_packages: true
--- a/sample_files/.travis.yml
+++ b/sample_files/.travis.yml
@@ -8,6 +8,8 @@ addons:
     packages:
       - expect-dev  # provides unbuffer utility
       - python-lxml  # because pip installation is slow
+      - python-simplejson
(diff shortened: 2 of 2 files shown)


bwrsandman:
I was hoping there were more packages we use in https://github.com/travis-ci/apt-package-whitelist/blob/master/ubuntu-precise

guewen:
:+1:

pedrobaeza:
:+1:

bwrsandman:
Comparing:
https://travis-ci.org/OCA/management-system/jobs/63392625
https://travis-ci.org/OCA/management-system/jobs/63487272

There is a saving of 20s in travis_install_nightly and an additional 0.07s in Installing APT Packages (BETA)

This is by no means a complete performance analysis, just one case. If we only look at these two cases, it's a pretty significant saving of time: -10% from travis_install_nightly!

pedrobaeza:
Of course every improvement is welcome!

bwrsandman:
List of other white list requests we could use:
- [ ] https://github.com/travis-ci/travis-ci/issues/3911
- [ ] https://github.com/travis-ci/travis-ci/issues/3901 (for runbot addons)
- [ ] https://github.com/travis-ci/travis-ci/issues/3734

PR #193, opened 2015-05-21
[IMP] Travis Transifex doc

pedrobaeza:


sample_files/.travis.yml
```diff
@@ -14,8 +14,10 @@ env:
   - VERSION="8.0" TESTS="0" LINT_CHECK="0" TRANSIFEX="0"
   - TRANSIFEX_USER='transbot@odoo-community.org'
   # This line contains the encrypted transifex password
-  # To encrypt transifex password, use
-  # $ travis encrypt TRANSIFEX_PASSWORD=your-password -r owner/project
+  # To encrypt transifex password, install travis ruby utils with:
+  #   $ [sudo] gem install travis
```
bwrsandman: The standard for displaying a sudo command is to use `#` as the prompt instead of `$`
Is it a good idea to suggest sudo?

Some might have travis (Trajectory Analyzer and Visualizer) accidentally installed. I know because I did.
Might be better to suggest installing it as the local user and calling it from where gem installs it.

pedrobaeza: I didn't get travis gem installed without sudo, because it directly install to gems directory. Is there a way to install it only for the user?

bwrsandman: Best doc I found is here[1]
If you add `--user-install` it will install it in `/home/username/.gem/ruby/1.9.1/bin`

[1] http://www.binarytides.com/quick-tip-installing-ruby-gems-in-the-users-home-directory/

pedrobaeza:
Put it gem user install.

bwrsandman:
:+1:

sample_files/.travis.yml
```diff
@@ -14,8 +14,10 @@ env:
   - VERSION="8.0" TESTS="0" LINT_CHECK="0" TRANSIFEX="0"
   - TRANSIFEX_USER='transbot@odoo-community.org'
   # This line contains the encrypted transifex password
-  # To encrypt transifex password, use
-  # $ travis encrypt TRANSIFEX_PASSWORD=your-password -r owner/project
+  # To encrypt transifex password, install travis ruby utils with:
+  #   $ gem install travis --user-install
+  # and use:
+  #   $ travis encrypt TRANSIFEX_PASSWORD=your-password -r owner/project
```
guewen: Can we suggest where to find the password or how to define it?

pedrobaeza: In case of OCA user, transbot, we should keep this user/password hidden for avoiding vandalism in the account.

guewen: @pedrobaeza I'm not talking about the OCA's transbot user, but about the token for transifex (`your-password` here).

guewen: And I don't propose to write them down here, but to explain how to find it or what is the procedure to get one.

bwrsandman: @max3903 could you chime in?

pedrobaeza: OK, we can expand it if you want

Issue #194, opened 2015-05-22
Activate Transifex on all projects


bwrsandman:
To activate a repo for transifex see #186 and https://github.com/OCA/maintainer-quality-tools/blob/master/sample_files/.travis.yml#L19

You need a test with `TRANSIFEX="1"`, `TRANSIFEX_USER='transbot@odoo-community.org'` and the secret depending on repo

I generated a secret for the transifex password for each repo:

``` python
repo_secures = {
    'OCA/OCB': 'OSzK/I+XGfKP+7L+yfzNemdPZBUncayvqmbm1+HqBFLvJs9VCIT/avdSAzVMSos8hgEzmzoXbLKxoRA8v+IOg20q6/l2UK4knO1hSPsr+iBFvz7Ou3lGeXAAulaMOtKN5b4Wj1LN8SbqjAN+ng0XyQ0CkxaKKM6hNyjux9bN7+I=',
    'OCA/account-analytic': 'CSf07Ubip0XXb19t6uBFhRMTlKFz7dGO2oxN5ibgObzjj+fncjdMsTqY0tIGT0/0+X9hWzDL4ZpY+1jhrHDgnDw9D2w81CZXC82DobPf+/pQ4MCHbZmldTqAE/APIzG76D6m77VB0yYGX+3LhNrpRKiDlWbLUFySQV5pe4py4R0=',
    'OCA/account-budgeting': 'QhZpJTRe5e/mw0gY0LPcPHM0kdR68sGEqaOnPaRuL44Ni04OZ4iZkVLybHdc+L7qKMN2rzFSkMqgpBA/B9zUasKC8P8ezdM8VU6ZgFt/5J364vERm27ys0CDMgpcuRlE8ciubMLsgh7PkTLNqG63AsZCEwObgvd0dT24HsCjxvA=',
    'OCA/account-closing': 'z1XJzAw3IBv62Jw8Piyb0pyHpiNDcivt9QoDeYFW0ZHH9BLlyjF09H3f6BGoW9SpK2T22EkSZyRJjjPpL+vVfKfpFalT7cOpWsFJyZwr7xxb96hYm6S/PcPPn0g3P/iKSyqv12s4GiumOJIG05Nq4XhcwiKvEzCHvUjHyUC59u8=',
[...]

PR #195, opened 2015-05-23
Transifex create options

bwrsandman:
Following @pedrobaeza's [suggestion](https://github.com/OCA/maintainer-quality-tools/issues/194#issuecomment-104869668), two more options to transifex creation have been adeed.

These can be overridden in the `.travis.yml` environment section.
* `TRANSIFEX_FILL_UP_RESOURCES`, default ~~False~~ True, activates TM at creation
* `TRANSIFEX_TEAM`, default [23907](https://www.transifex.com/organization/oca/team/23907/), joins the OCA team at creation instead of creating another

~~The default for `TRANSIFEX_FILL_UP_RESOURCES` is up for debate, I will change it to True if enough people want it. Personally I would rather not have auto fill of terms to 100% if they're done automatically.~~

### DIFF
--- a/sample_files/.travis.yml
+++ b/sample_files/.travis.yml
@@ -21,9 +21,11 @@ env:
   # The default project slug is owner-repo_name-version (with dash in the version string).
   # The default organization is the owner of the repo.
   # The default fill up resources (TM) is True.
+  # The default team is 23907. https://www.transifex.com/organization/oca/team/23907/
   # - TRANSIFEX_PROJECT_SLUG=
   # - TRANSIFEX_ORGANIZATION=
   # - TRANSIFEX_FILL_UP_RESOURCES=
+  # - TRANSIFEX_TEAM=
 
   matrix:
   - LINT_CHECK="1"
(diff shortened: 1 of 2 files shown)


bwrsandman:
Another thing. I haven't tested this on any project yet. So don't merge until it has been tried

pedrobaeza:
I think TM is not problem, because the match is only done when matching 100%. Odoo is using it, and they get 10% of terms automatically filled: https://twitter.com/odony/status/601040459650109441

bwrsandman:
@pedrobaeza could you test this branch and see if it works?

pedrobaeza:
I'm preparing l10n-spain 7.0 branch for it :wink:

pedrobaeza:
OK, in 39 suggested terms I have checked, only 3 were wrong: two by gender switch (but that's a common problem from english to spanish - for example, canceled can be translated as "Cancelado" or "Cancelada"), and another one that was originally bad translated, but it's understandable enough: "Refund" > "Rectificado" instead of "Rectificativa".

So for me it's OK to activate it. Anyways, all suggested terms are put on unreviewed terms, so you can check it later to know which ones are correct.

pedrobaeza:
Another thing, for this PR, please document these two variables on the .travis.yml sample.

bwrsandman:
Done

pedrobaeza:
:+1:

Ehtaga:
:+1:

guewen:
:+1:

Issue #35, opened 2014-07-16
Fix test coverage measurement


dreispt:
The current use of `coveralls` used in PR #29 is meaningless for shell scripts.
Would it be better to remove it?
Or perhaps add shell test coverage checking using [shcov](https://code.google.com/p/shcov/) or similar.

bwrsandman:
We have converted a lot of the scripts to python, I think now is a good time to re-activate coveralls

moylop260:
@dreispt 
Can we close this issue?

PR #197, opened 2015-05-27
Reactivate python coverage on MQT itself

bwrsandman:
The point in #35 is no longer the case since most tests are python.

Improve coverage of getaddons.py and bring overall coverage to about 90%

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,0 +1,8 @@
+[report]
+# Regexes for lines to exclude from consideration
+exclude_lines =
+    # Have to re-enable the standard pragma
+    pragma: no cover
+
+    # Don't complain if non-runnable code isn't run:
+    if __name__ == .__main__.:
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,3 +1,6 @@
+[run]
+data_file = .coverage_mqt  # to avoid being overridden by server tests
+
 [report]
 # Regexes for lines to exclude from consideration
 exclude_lines =
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,5 +1,6 @@
 [run]
 data_file = .coverage_mqt  # to avoid being overridden by server tests
+omit = *system_site_packages*
 
 [report]
 # Regexes for lines to exclude from consideration
--- a/.travis.yml
+++ b/.travis.yml
@@ -40,8 +40,8 @@ install:
   - travis_install_nightly 8.0 # only used if VERSION not set in env
 
 script:
-  - coverage run ./travis/self_tests
-  - coverage run ./travis/travis_run_tests 8.0  # only used if VERSION not set in env
+  - coverage run --append ./travis/self_tests
+  - coverage run --append ./travis/travis_run_tests 8.0  # only used if VERSION not set in env
 
 after_success:
   - coveralls
--- a/.travis.yml
+++ b/.travis.yml
@@ -40,8 +40,8 @@ install:
   - travis_install_nightly 8.0 # only used if VERSION not set in env
 
 script:
-  - self_tests
-  - travis_run_tests 8.0  # only used if VERSION not set in env
+  - coverage run ./travis/self_tests
+  - coverage run ./travis/travis_run_tests 8.0  # only used if VERSION not set in env
 
-#after_success:
-#  coveralls
+after_success:
+  - coveralls
--- a/travis/self_tests
+++ b/travis/self_tests
@@ -1,12 +1,25 @@
 #!/usr/bin/env python
 
 import os
+import getaddons
 from test_server import get_test_dependencies
 
 
-repo_dir = os.environ.get("TRAVIS_BUILD_DIR", "../tests/test_repo/")
+repo_dir = os.environ.get("TRAVIS_BUILD_DIR", "./tests/test_repo/")
+exclude = os.environ.get("EXCLUDE")
+
 addons_list = ['test_module', 'second_module']
 to_preinstall = get_test_dependencies(repo_dir, addons_list)
 
 assert not [x for x in to_preinstall if x in addons_list], \
     "Should not preinstall modules to test!"
+
+# Testing getaddons
+assert getaddons.main() == 1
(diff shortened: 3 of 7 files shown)


pedrobaeza:
:+1:

yvaucher:
:+1:

bwrsandman:
Not sure why coverage is only done for the linter...

travis/travis_run_tests
```diff
@@ -10,13 +10,30 @@ from travis_helpers import success_msg, fail_msg
...
     for test in test_list:
         # keep backward compatibility with version as an argument
         print("======== Testing %s ========" % test[0])
         test_w_args = test + args
-        res = subprocess.call(test_w_args)
+        test_file = test_w_args[0]
+        if test_file.endswith(".py"):
+            test_lib = test_file[:-3]
+            try:
+                res = __import__(test_lib).main(argv=test_w_args)
```
dreispt: This adds an implicit requirement that `.py` test files must have a `main()` function.
I don't see that as an issue but we should explicitly explain that somewhere (in comment code or in a README).

dreispt:
Small comment but looks great! :+1:

gurneyalex:
:+1:

travis/travis_run_tests
bwrsandman: How about in the docstring?

bwrsandman:
Still not getting any files in coverage for tests other than the linter and getting source code not available on coveralls:
https://coveralls.io/files/786085237

bwrsandman:
Is it possible that the coverage call in test server deletes the coverage cache from these tests?

bwrsandman:
Worked with 22edf3c, but let's try without changing the coverage cache

bwrsandman:
Not changing the cache makes it so only the linter shows coverage.

.travis.yml
```diff
@@ -40,8 +40,8 @@ install:
   - travis_install_nightly 8.0 # only used if VERSION not set in env
 
 script:
-  - self_tests
+  - coverage run --rcfile=.coveragerc_mqt ./travis/self_tests
   - travis_run_tests 8.0  # only used if VERSION not set in env
```
bwrsandman: @dreispt Should I cover this line?

I need an absolute path, I can't use `$PATH` directly because of the `cp` in the setup

If I cover it, I need to change the line to :
```
- coverage run --rcfile=.coveragerc_mqt ./travis/travis_run_tests 8.0  # only used if VERSION not set in env
```

Currently, `travis_run_tests` is in `$HOME` not `$HOME/travis`, the coveralls won't work at associating the files like that.

bwrsandman: I kind of need to cover that line, self_tests doesn't cover a lot.

Keep that in mind for your review

bwrsandman:
\o/ It's working! 

Please note that I am directly invoking the scripts and not using the `$PATH`

travis/self_tests
```diff
@@ -1,12 +1,25 @@
...
+
+# Testing getaddons
+assert getaddons.main() == 1
+getaddons.main(["getaddons.py", repo_dir])
+getaddons.main(["getaddons.py", "-m", repo_dir])
+getaddons.main(["getaddons.py", "-m",
+                repo_dir + "/" if repo_dir[-1] == '/' else repo_dir])
+if exclude:
+    getaddons.main(["getaddons.py", "-m", repo_dir, "-e", exclude])
+    getaddons.main(["getaddons.py", "-e", exclude, repo_dir])
```
dreispt: Maybe I'm missing something, but shouldn't we have some `assert`s somewhere to check the results?

bwrsandman: The results are printed in stderr. Can't really do that here. But this isn't checking for that, it's just running and checking for a raise.

.coveragerc
```diff
@@ -0,0 +1,12 @@
...
+omit = *system_site_packages*
+
+[report]
+# Regexes for lines to exclude from consideration
+exclude_lines =
+    # Have to re-enable the standard pragma
+    pragma: no cover
+
+    # Don't complain if non-runnable code isn't run:
+    if __name__ == .__main__.:
```
bwrsandman: Anyone know why this isn't being excluded? https://coveralls.io/builds/3173483/source?filename=travis%2Fgetaddons.py#L77

travis/self_tests
dreispt: OK

dreispt:
:+1:

.coveragerc
gurneyalex: @bwrsandman because the content of the `if` block you exclude is not matched by the regexp? (only the `if` line itself)

bwrsandman: The if line isn't being matched either and an excluded if line would ignore the block

.travis.yml
```diff
@@ -40,8 +40,8 @@ install:
...
 script:
-  - self_tests
-  - travis_run_tests 8.0  # only used if VERSION not set in env
+  - coverage run --append ./travis/self_tests
+  - coverage run --append ./travis/travis_run_tests 8.0  # only used if VERSION not set in env
 
-#after_success:
-#  coveralls
+after_success:
+  - coveralls --rcfile=.coveragerc_mqt
```
bwrsandman: Oops, this shouldn't be

.coveragerc
bwrsandman: The issue is in .travis.yml, I left in a custom .coveragerc_mqt reference

bwrsandman:
@dreispt could you take a look at the coverage and see if the not run lines make sense?
https://coveralls.io/builds/3173483/source?filename=travis%2Ftravis_run_tests

dreispt:
We need the except block fro the nice results summary to be printed out. We'll have to figure a way to test those lines.

Running a non-python test is not used but I would avoid removing the feature, and adding a simple test for this is not hard.

I say merge it as it is: you already did a great job and we can improve those self tests later with another PR.

bwrsandman:
Thanks! I will await merges then.

dreispt:
I'm counting four thumbs up, so merging...

PR #201, opened 2015-06-02
add code climate sample

lepistone:


sample_files/.codeclimate.yml
```diff
@@ -0,0 +1,7 @@
+languages:
+  JavaScript: true
+  Python: true
+exclude_paths:
+- "__unported__/*"
+- "*/__openerp__.py"
+- "*/migrations/*"
```
pedrobaeza: migrations can be interesting to test

pedrobaeza:
Do we remove `__init__.py` also?

bwrsandman:
@pedrobaeza `__init__.py` don't gave bad grades generally. It would be important to see if there are issues for odd `__init__.py` files

`__openerp__.py` files have been ignored due to the fact that they're mostly duplicate code and code climate suggests you refactor or make a class.

pedrobaeza:
:+1:

bwrsandman:
@lepistone Could you add comments to the sample file to explain why something is ignored. This doesn't need to go in the maintainer tools script, but here it would be nice to explain the idea behind it.

*  "__unported__/*": obvious, no need for comment
* "*/__openerp__.py": No choice for duplicated code, this file is fed in an `eval`
* "*/migrations/*": Without OpenUpgrade in pip, the functions need to be copied for each migration file

lepistone:
@bwrsandman done

bwrsandman:
Could you add a sample badge to sample_files/README.md

lepistone:
@bwrsandman done

bwrsandman:
:+1:

Issue #206, opened 2015-06-15
Test coverage and module initialization code


dreispt:
Module initialization methods are run when the module is first installed, and not when it is tested.
This makes that code to be marked as not covered ([example](https://coveralls.io/builds/2817711/source?filename=project_stage_state%2Fproject.py#L38)).

To get it covered We could add tests call it, but sometimes this init code shouldn't be run twice.
What would be the best way do deal with this?

bwrsandman:
Perhaps a mock, or perhaps a custom setup and rollback case.

dreispt:
I have an idea: detect and preinstall all core and external dependencies, and then perform the tests on installation (`-i`) of the repo modules to test.

PR #212, opened 2015-07-15
Preinstall dependencies and run tests with --init instead of --upgrade

dreispt:
Currently the database setup phase install all modules, and tests are run with a second command, using an upgrade.

This has a couple of issues:
  * In some cases tests during an upgrade have different results from tests during and install.
  * Init hooks don't run during an upgrade and their code is reported as not being covered by tests.

This PR changes the tests, so that only the core an external module dependencies are preinstalled in the setup phase, so the repo module tests can be executed with a `--init` instead of `--update`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -38,6 +38,7 @@ install:
   - travis_install_nightly 8.0 # only used if VERSION not set in env
 
 script:
+  - self_tests
   - travis_run_tests 8.0  # only used if VERSION not set in env
 
 #after_success:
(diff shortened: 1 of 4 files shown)


dreispt:
Just made a fix to make sure modules to test are not preinstalled and added unit tests for that logic.

dreispt:
Fixes #206

gurneyalex:
@dreispt do you have an idea of the impact on our branches?

gurneyalex:
:+1:

hbrunn:
:+1:

lmignon:
:+1:

PR #214, opened 2015-07-22
fix #213: handle all lint/test cases, refactor

### LINKED ISSUE #213: tests are not being executed by default since #186
See for example the build
- https://travis-ci.org/OCA/purchase-workflow/builds/68473009

the `.travis.yml` does not specify `LINT_CHECK` nor `TESTS`. Since the merge of #186, the default becomes `LINT_CHECK = 1` and `TESTS = 1`, and in that case, surprisingly, no tests are run.

There is probably a quick fix changing that line: 
https://github.com/OCA/maintainer-quality-tools/blob/master/travis/travis_run_tests#L48

We could do that rapidly, I can make a PR for that.

On the other hand, I find that logic a bit confusing because we try to handle too many cases (one action per line, or many) in a very complex way.

- with all defaults to one, the configuration becomes backwards and difficult to follow (for me at least). Specifically, every time we do `SOMETHING = 1` it is actually a no-op
- what actually gets done is not very explicit, because flags interact which each other in ways I find difficult to understand.

Suggestion 1: remove the defaults to one, and remove TESTS altogether (tests are run if there are no other options) so that old repos with a trivial config like purchase-workflow 7.0 still run the tests (no lint, but that is less harmful). example:
```yaml
- 8.0
- 8.0 LINT_CHECK=1
- 8.0 TRANSIFEX=1
```

Suggestion 2: make everything explicit, with no default, (but run everything where there are no variables at all for compatibility)
```yaml
- 8.0 TESTS=1
- 8.0 LINT_CHECK=1
- 8.0 TRANSIFEX=1
```

Thanks to all for your time!

lepistone:
fix #213: no tests only with explicit LINT_CHECK

We are trying to handle many different situations here:

- a .travis.yml that always sets variables to 0 or 1, like the one in
  samples/.travis.yml (4 cases)
- a line that has no options at all (it should lint, then test)
- a line that has only LINT_CHECK=1 (special case: it should not run
  tests)
- a line that has only LINT_CHECK=0 (special case: it should only run
  tests)

That makes a total of 7 cases that should all be handled here.

See #213 for a discussion on how to make all that more explicit (will
require an update in all .travis.yml files).

### DIFF
--- a/travis/travis_run_tests
+++ b/travis/travis_run_tests
@@ -50,7 +50,7 @@ if __name__ == '__main__':
     if tests_unspecified and not lint_check_enabled:
         tests.append(['test_server.py'])
 
-    if tests_enabled:
+    elif tests_enabled:
         tests.append(['test_server.py'])
 
     if transifex_enabled and not is_pull_request:
--- a/travis/travis_run_tests
+++ b/travis/travis_run_tests
@@ -31,24 +31,31 @@ def main(test_list):
 
 
 if __name__ == '__main__':
-    do_lint_check = os.environ.get("LINT_CHECK", '1') == "1"
-    do_tests = os.environ.get("TESTS", '1') == "1"
-    export_transifex = os.environ.get("TRANSIFEX") == "1"
-    # Variable TRAVIS_PULL_REQUEST corresponds to pull request number if the
-    # current job is a pull request, "false" if it's not a pull request.
-    travis_pull_request = os.environ.get("TRAVIS_PULL_REQUEST")
+    lint_check_disabled = os.environ.get('LINT_CHECK') == '0'
+    lint_check_enabled = os.environ.get('LINT_CHECK') == '1'
+    tests_enabled = os.environ.get('TESTS') == '1'
+    tests_disabled = os.environ.get('TESTS') == '0'
+    tests_unspecified = os.environ.get('TESTS') is None
+    travis_enabled = os.environ.get('TRANSIFEX') == '1'
+    transifex_enabled = os.environ.get('TRANSIFEX') == '1'
+
+    # TRAVIS_PULL_REQUEST contains the pull request number or 'false'
+    is_pull_request = os.environ.get('TRAVIS_PULL_REQUEST') != 'false'
 
     # Test list. Each test is a list with command + arguments.
     tests = []
-    # By default, do lint checks
-    if do_lint_check:
+
+    if not lint_check_disabled:
         tests.append(['test_flake8'])
         tests.append(['test_pylint'])
-    # Execute tests by default. Disable them on lint checks
-    if not do_lint_check and do_tests:
+
+    if tests_unspecified and not lint_check_enabled:
         tests.append(['test_server.py'])
-    # Execute transifex export with no tests and if it's an official commit
-    if not tests and export_transifex and travis_pull_request == "false":
+
+    if tests_enabled:
+        tests.append(['test_server.py'])
+
+    if transifex_enabled and not tests_disabled and not is_pull_request:
         tests.append(['travis_transifex.py'])
     if tests:
         exit(main(tests))
--- a/travis/travis_run_tests
+++ b/travis/travis_run_tests
@@ -55,7 +55,8 @@ if __name__ == '__main__':
     if tests_enabled:
         tests.append(['test_server.py'])
 
-    if transifex_enabled and not tests_disabled and not is_pull_request:
+    if transifex_enabled and not is_pull_request:
         tests.append(['travis_transifex.py'])
+
     if tests:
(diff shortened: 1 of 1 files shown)


travis/travis_run_tests
```diff
@@ -44,8 +44,8 @@ if __name__ == '__main__':
     if do_lint_check:
         tests.append(['test_flake8'])
         tests.append(['test_pylint'])
-    # Execute tests by default. Disable them on lint checks
-    if not do_lint_check and do_tests:
+    # Disable tests if LINT_CHECK is not mentioned explicitly.
```
gurneyalex: I suggest:
s/mentioned/enabled/

gurneyalex:
minor rewording suggested, :+1:

dreispt:
Having bith `TESTS="1" LINT_CHECK="1"` should work and do both...

lepistone:
Seems like you're right @dreispt. My brain just exploded.

lepistone:
I updated the fix to handle more cases. See the updated commit message (or the updated title / description of this pull request, which is the same).

Thanks!

travis/travis_run_tests
```diff
@@ -31,24 +31,31 @@ def main(test_list):
...
+
+    if tests_unspecified and not lint_check_enabled:
         tests.append(['test_server.py'])
-    # Execute transifex export with no tests and if it's an official commit
-    if not tests and export_transifex and travis_pull_request == "false":
+
+    if tests_enabled:
+        tests.append(['test_server.py'])
+
+    if transifex_enabled and not tests_disabled and not is_pull_request:
```
dreispt: Does Transifex require test to run?
`not tests_disabled` is not equivalent to `not tests` in the prior code...
The comment is helpful, you should keep it.

lepistone: You're right: since transifex already needs to be enabled explicitly, there is no need to complicate the logic looking into `TESTS`. I fixed that.

As for the comments, I refactored the decisions so that the python code corresponds to the  old comments. At that point they are not useful anymore (and I start to understand something).

Thanks!

dreispt:
Looks good :+1:

travis/travis_run_tests
```diff
@@ -31,24 +31,32 @@ def main(test_list):
...
-    export_transifex = os.environ.get("TRANSIFEX") == "1"
-    # Variable TRAVIS_PULL_REQUEST corresponds to pull request number if the
-    # current job is a pull request, "false" if it's not a pull request.
-    travis_pull_request = os.environ.get("TRAVIS_PULL_REQUEST")
+    lint_check_disabled = os.environ.get('LINT_CHECK') == '0'
+    lint_check_enabled = os.environ.get('LINT_CHECK') == '1'
+    tests_enabled = os.environ.get('TESTS') == '1'
+    tests_disabled = os.environ.get('TESTS') == '0'
+    tests_unspecified = os.environ.get('TESTS') is None
+    travis_enabled = os.environ.get('TRANSIFEX') == '1'
```
Ehtaga: Why `travis_enabled` ? You already have `transifex_enabled` to deal with `TRANSIFEX`.
And `tests_disabled` is not used so unnecessary for the moment.

Ehtaga:
By using `os.environ.get("TESTS", '1')`, we set a default value to `do_tests`, so you couldn't have the case of unspecified tests.
The problem was this line actually: `if not do_lint_check and do_tests:` because when nothing is specified for `TESTS` and for `LINT_CHECK`, both are set to default value (1) so tests were never triggered.

travis/travis_run_tests
lepistone: Thanks. I was expecting the linter to catch those, but it doesn't.

lepistone:
@Ehtaga the first fix proposed in this PR was indeed a change of that line only. However it didn't handle all the 7 cases we seem to need.

Thanks!

travis/travis_run_tests
```diff
@@ -31,24 +31,30 @@ def main(test_list):
 
 
 if __name__ == '__main__':
-    do_lint_check = os.environ.get("LINT_CHECK", '1') == "1"
-    do_tests = os.environ.get("TESTS", '1') == "1"
-    export_transifex = os.environ.get("TRANSIFEX") == "1"
-    # Variable TRAVIS_PULL_REQUEST corresponds to pull request number if the
-    # current job is a pull request, "false" if it's not a pull request.
-    travis_pull_request = os.environ.get("TRAVIS_PULL_REQUEST")
+    lint_check_disabled = os.environ.get('LINT_CHECK') == '0'
```
moylop260: Please, consider the case: LINT_CHECK unset

moylop260:
:+1:

gurneyalex:
:+1:

PR #216, opened 2015-07-28
Remove E1124 (redundant-keyword-arg) from lint

bwrsandman:
Arguments can benefit from having keywords, especially in python.

Sometimes arguments aren't clear in a non-typed language like python. In this case, we can write self documenting code by using keywords.

See practical example here: https://github.com/OCA/account-fiscal-rule/pull/27/files#diff-473eef115068a06e3800748db81bb2beR69

```python
# Unclear
self._fiscal_position_map(
    # ...
    address_data['delivery'],
    # ...
)
# Much more clear
self._fiscal_position_map(
    # ...
    partner_shipping_id=address_data['delivery'],
    # ...
)
```


dreispt:
:+1:

pedrobaeza:
:+1:

yvaucher:
This error seems to be trigged because pylint doesn't know the effect of the decorator.
Thus it thinks partner_shipping_id is given twice. But first variable are consumed by cr and uid.

Maybe you could bypass this by doing cr=cr and uid=uid or by porting your onchange to new api using decorator @api.multi (not @api.onchange)

This warning could be convenient to warn when you forget to remove a self from a method call. (If I understand correctly this warning message)

:-1:

pedrobaeza:
What I understand of this argument is not that, @yvaucher. I think it's that you shouldn't put the argument name in the passed arguments if the argument is not a keyword. So I see this not useful to prevent it.

bwrsandman:
@yvaucher That's not it. It will do this without a decorator. This check looks for keywords being used when they don't need to.

yvaucher:
redundant-keyword-arg (E1124):
 	    Argument %r passed by position and keyword in %s call Used when a function call would result in assigning multiple values to a function parameter, one value from a positional argument and one from a keyword argument.

yvaucher:
By the way in you example, 

    @api.model
    def _fiscal_position_map(
        self, result, partner_id=None, partner_invoice_id=None,
        partner_shipping_id=None, company_id=None
    ):

Here all raising E1124 attributes are keywords and with all the context=context we put everywhere and all the super(...).method passing kwargs by name, I seriously doubt the message is what you understood.

For instance, when reviewing, I strongly suggest to propagate kargs using the key which is the good way to go for clarity as you said. You might solve you issue by adding result=result in method call (but pylint might tell there are extra parameters). But I would prefer to see that method ported to 8.0 with `@api.multi`

moylop260:
I will close this PR. if is a future problem again we could open it again.

PR #223, opened 2015-08-05
[REF] test_pylint: Enable C0103 to validate CamelCase class name. (For old version snake_case is allow it too)

moylop260:
Reviewers comment PR that we need to use CamelCase for 8.0 version and we have a [guideline](https://github.com/OCA/maintainer-tools/blob/master/CONTRIBUTING.md#odoo-python-class)

This PR enable this case.

Currently pylint and flake8 don't have `expected error`, I will to work in this other feature in other PR to add test of lint cases.

By now, I share next image with testcases:
captura de pantalla 2015-08-05 a las 6 47 52 p m

And I enabled this PR in .travis.yml in next change:
https://github.com/Vauxoo/rma/pull/43
(Red because this branch has snake_case class name)

FYI we have a script to auto-fix this case in py files more info [here](https://github.com/Vauxoo/pylint-conf/tree/master/autopep8_extended)

NOTE: The pylint fail named `C0103` enable validation for: **class**, module, constant, function, method, attributes, argument and variable names.
This PR just enable `class name` with regular expression CamelCase in pylint.conf file other checks use a regex `.*`
NOTE2: For old version `class-name-regex` is overwriten of pylint.conf with parameter in `test_pylint` script to allow `snake_case` too.

### DIFF
--- a/travis/test_flake8
+++ b/travis/test_flake8
@@ -6,4 +6,11 @@ flake8 . --config=${FLAKE8_CONFIG_DIR}/travis_run_flake8__init__.cfg
 status1=$?
 flake8 . --config=${FLAKE8_CONFIG_DIR}/travis_run_flake8.cfg
 status2=$?
-exit $((${status1} || ${status2}))
+
+if [[ "${status1}${status2}" == "00" ]]; then
+  echo "Flake8 OK"
+  exit 0
+else
+  echo "Flake8 ERRORS"
+  exit 1
+fi
--- a/travis/test_pylint
+++ b/travis/test_pylint
@@ -17,7 +17,13 @@ fi
 IFS="/" read -a REPO <<< "${ODOO_REPO}"
 export PATH=${PATH}:${PWD}/../${REPO[1]}-${VERSION}/openerp
 
+if [[ "${VERSION}" == 6.* ]] || [[ "${VERSION}" == 7.* ]] ; then
+    # snake_case and CamelCase allow it for old version
+    export CLASS_CASE='--class-rgx=([a-z_][a-z0-9_]{2,45})|([A-Z_][a-zA-Z0-9]{2,45})$'
+fi
+
+
 #run pylint command
-pylint --rcfile=${PYLINT_CONFIG_DIR}/travis_run_pylint.cfg ${MODULES_TO_TEST}
+pylint --rcfile=${PYLINT_CONFIG_DIR}/travis_run_pylint.cfg ${CLASS_CASE} ${MODULES_TO_TEST}
 pylint_status=$?
 exit $((${pylint_status}))
--- a/travis/test_pylint
+++ b/travis/test_pylint
@@ -22,8 +22,13 @@ if [[ "${VERSION}" == 6.* ]] || [[ "${VERSION}" == 7.* ]] ; then
     export CLASS_CASE='--class-rgx=([a-z_][a-z0-9_]{2,45})|([A-Z_][a-zA-Z0-9]{2,45})$'
 fi
 
-
 #run pylint command
 pylint --rcfile=${PYLINT_CONFIG_DIR}/travis_run_pylint.cfg ${CLASS_CASE} ${MODULES_TO_TEST}
 pylint_status=$?
-exit $((${pylint_status}))
+if [[ "$pylint_status" == "0" ]]; then
+  echo "Pylint OK"
+  exit 0
+else
+  echo "Pylint ERRORS"
+  exit 1
+fi
--- a/travis/travis_install_nightly
+++ b/travis/travis_install_nightly
@@ -2,11 +2,6 @@
 
 pip install -q QUnitSuite flake8 coveralls pylint > /dev/null 2>&1
 
-# We can exit here and do nothing if this only a LINT check
-if [ "${LINT_CHECK}" == "1" ] ; then
-    exit 0
-fi
-
 # For backward compatibility, take version from parameter if it's not globally set
 if [ "x${VERSION}" == "x" ] ; then
     VERSION="${1}"
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,22 +26,24 @@ env:
   - VERSION="8.0" TESTS="1" LINT_CHECK="0"
 
   matrix:
-  - INCLUDE="broken_module" SERVER_EXPECTED_ERRORS="1"  # test errors are detected
+  # Test for: YAML assert error; test snake case lint
+  - INCLUDE="broken_module,broken_snake_case" LINT_CHECK="1" EXPECTED_ERRORS="2"
   - EXCLUDE="broken_module" OPTIONS="--log-level=debug" INSTALL_OPTIONS="--log-level=info"
   - INCLUDE="test_module,second_module" UNIT_TEST="1"
-  - VERSION="7.0" INCLUDE="test_module,second_module" ODOO_REPO="OCA/OCB"  # ODOO_REPO usage example
+  # Tests for: version 7; OCB usage; snake case check not done for v7
+  - VERSION="7.0" INCLUDE="test_module,second_module,broken_snake_case" LINT_CHECK="1" ODOO_REPO="OCA/OCB"
   - VERSION="6.1" INCLUDE="test_module,second_module"
-  - LINT_CHECK="1" TESTS="0"
+  - LINT_CHECK="1" TESTS="0" EXPECTED_ERRORS="1"  # Lint error expected
 
 install:
   - cp -r ../maintainer-quality-tools/ $HOME
   - mv tests/test_repo/* ./
   - export PATH=$HOME/maintainer-quality-tools/travis:$PATH
-  - travis_install_nightly 8.0 # only used if VERSION not set in env
+  - travis_install_nightly
 
 script:
   - coverage run --append ./travis/self_tests
-  - coverage run --append ./travis/travis_run_tests 8.0  # only used if VERSION not set in env
+  - coverage run --append ./travis/travis_run_tests
 
 after_success:
   - coveralls
--- a/git/README.md
+++ b/git/README.md
@@ -28,3 +28,7 @@ sudo pip install pylint --upgrade
 
 You can bypass these checks setting environment variable NOLINT before calling
 commit, e.g, `NOLINT=1 git commit`.
+
+You can use a lint configuration setting environment variable VERSION with
+the number of version of odoo before calling
+commit, e.g, `VERSION=7.0 git commit`.
--- a/git/pre-commit
+++ b/git/pre-commit
@@ -21,8 +21,14 @@ GIT_DIR="$(dirname $(dirname $(dirname $0)))"
 export PYTHONPATH=${PYTHONPATH}:$GIT_DIR
 touch $GIT_DIR/__init__.py
 
+if [[ "${VERSION}" == 6.* ]] || [[ "${VERSION}" == 7.* ]] ; then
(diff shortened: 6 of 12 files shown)


nhomar:
:+1:

In our case we see at least 2 or 3 comments regarding things "That Pylint can do" being done manually on PR which represent at least 2 minutos between what reviewer say something and contributor aprove or make the change if this rules are automates we can save such time because travis will inform them before the PR start to discuss and save a lot of time:

20 minuites with 20 or 30 persons contributing all days Make your numbers.

Regards.

pedrobaeza:
You have to check the own Travis. It is also needed to change git pre-commit hook in the git directory.

nhomar:
@pedrobaeza 

There is 3 different things.

***1 Thing*** is "How do we avoid a mistake", for that the prehook is the tool you use, (we simply run manually but it is another different history).

***2 Thing*** The rule itself, I can not check something locally that I do not considere an error (for example the CamelCase) for that is the configuration proposed here, It is really time consuming and tired expend our precious time commenting that "Somebody need CamelCase" it this will be a rule, then let's set the rule and stop avoiding it, from time to time we will fix all repositories (for that an script will come in some time from us always and here is the third thing).

***3rd Thing*** the automation (out of scope of this PR), if I am migrating from "Private Repositories or personals ones" to "OCA" it do not make sense that people expend hours reading comments, guidelines and then the Point of view of reviewers on specific and simple things like this one... "Use CamelCase" IMHO If it is a rule -good- we follow the rule and we have an script that simple will find the class and its declarations and change them in a Click. (we do that all the time for all internal rules).

Then resuming.

If the Topic CamelCAse is a rule, this PR must be merged to measure the rule, if it is an "optional rule" pleaseeee we clos happily this PR and for god sake stop ask to contributors such topic, let's them do what ever they want.

Reflexion:

We as reviewers shoudl automate as much LINT check as we think based on whatever rule we think using the correct tools (pylint for example) in order to affect with only 1 works 100% of repositories.

And the manual Job the "Code per review" must be done for the functionality itself, the quality of the view the crossed references and quality of the algorithms, semantic typo errors etc, etc, etc.... not to say somebody else "Use CamelCase", for me the rule should be:

>>"IF pylint complains then it is wrong if not it is subjective Propose the change of the rule globally in the channels dedicated for that and wait for such approval, the code that you will do will have -your rule- respect others work, time and opinions."

This statement is "Automatable" and "Measurable" I really do not enjoy at all discussing about PEP or NOT PEP, if it is PEP and Pylint pass silently... "GOOD" if not "BAD".

>> What about the subjective rules? we can discuss first all of them, but odoo's guideline and programming good practices (with the respective fundament that support each statement) should be done.

Example: sum strings versus concatenate properly with format versus concatenate with tuples.... each one of them can be discussed, agumented and managable with each technical impact they are not 1-0 rules, ther is cases where each one of them is good and enough for.

Thanks for the feedback.

PS: BTW your statement for me personally sound like "You are stupid you do not know GIT", and IMHO as one of the responsibles of this project I feel totally different I feel like : "I am stupid why didn't I automate that" that's why We are doing this PR.

pedrobaeza:
Please add here the same check: https://github.com/OCA/maintainer-quality-tools/blob/master/git/pre-commit#L25. But maybe we will find here a problem, because not all configuration is correctly get from the configuration file. Let's try.

moylop260:
@pedrobaeza 
I added the [change](https://github.com/OCA/maintainer-quality-tools/pull/223/files#diff-41534b470a0a330c9c125702b9967f11R25) in git hooks

﻿And I tested in local fine!
captura de pantalla 2015-08-06 a las 9 59 54 a m

pedrobaeza:
Thanks for the changes, @moylop260.

I have tried the git pre-commit script, and the problem enabling this directive is that we get other errors that aren't class names:

```
[C0103(invalid-name), ...] Invalid variable name "ok"
[C0103(invalid-name), ...] Invalid variable name "dt"
[C0103(invalid-name), ...] Invalid method name "_prepare_order_line_invoice_line"
```

and this is going to get to red all the branches for sure. Can we isolate the check only for class names?

nhomar:
2015-08-07 20:05 GMT-05:00 Pedro M. Baeza <notifications@github.com>:

> I have tried the git pre-commit script, and the problem enabling this
> directive is that we get other errors that aren't class names:
>
> [C0103(invalid-name), ...] Invalid variable name "ok"
> [C0103(invalid-name), ...] Invalid variable name "dt"
> [C0103(invalid-name), ...] Invalid method name "_prepare_order_line_invoice_line"
>
>
But In my opinion and the official doc for pylint[1] "ok" name is incorrect
(2 letters with an spelling world) dt is incorrect alco because only 2
letters, and the method I asume the log name is incorrect.

It is fine no? pylint should fail because of that I think? if not we should
add such role as ignoring it failing in pylint / pep itself.


We can exclude dt for example (as we did with cr but "ok" is not the same
case I think).

[1]http://pylint-messages.wikidot.com/messages:c0103

-- 
--------------------
Saludos Cordiales

[image: --]
Nhomar Hernandez

nhomar:
I confirm (I didn't check before);

Short names of variables between 2 and 30,

ok: has 2 letters, I am agreed ok fails.

dt: has 2 letters I am agreed it staty because it is common to alias datetime (frequently in modules) with such name (like cr and id).

method long: len('_prepare_order_line_invoice_line') = 32 : it should have 30 can you short your method name? or you can in the line (if you ar not agreed) add an explicit inline ignore for the method:

# disable=C0103

- We should validte all pylint (in my opinion) not only class names and "whateverwethinkigood" not.

What do you think!?

pedrobaeza:
The problem here is that if we include now this directive, we will need again to clean all OCA branches, because there is such variable names for sure across all code, so it's a difficult task to make.

The third one also puzzles me why is incorrect.

nhomar:
The third one is because it has >30 characters in the name of a method which is incorrect in pylint.

We can change it no poblem just define the long. (I like 30)

moylop260:
@pedrobaeza 
Can you give the line of code or repo where you saw this output?

pedrobaeza:
That was a test I made with a personal repository with git pre-commit hook (not with Travis one). Maybe is there any difference?

moylop260:
@pedrobaeza 
I used your same sentence but I can't to reproduce with git hook
My example [here](https://asciinema.org/a/24737)

Can you verify that you have `pre-commit` file and `conf` folder too.

moylop260:
@pedrobaeza 
I created changes after of PR
Can you check if you have next SHA:
https://github.com/vauxoo-dev/maintainer-quality-tools/commit/e56040cb9ccca9a9456ff1d94294dff74de7019f

moylop260:
FYI we have created a new script to replace snake_case to CamelCase
https://github.com/Vauxoo/pylint-conf/tree/master/autopep8_extended
Can you test it, please?
`autopep8_extended.py --select CW0001 -i YOUR_CODE_PATH`

dreispt:
@moylop I made a PR adding tests: https://github.com/vauxoo-dev/maintainer-quality-tools/pull/1

dreispt:
With the tests :+1:

moylop260:
@dreispt 
Good IMP!
Thank you!

moylop260:
==offtopic==
Where is good to push the script to auto-change snake to CamelCase of [autopep8_extended](https://github.com/Vauxoo/pylint-conf/tree/master/autopep8_extended)?

dreispt:
Sounds like a tool for maintainers: I think it's place is in `OCA/maintainer-tools`

travis/travis_install_nightly
```diff
@@ -2,11 +2,6 @@
 
 pip install -q QUnitSuite flake8 coveralls pylint > /dev/null 2>&1
 
-# We can exit here and do nothing if this only a LINT check
-if [ "${LINT_CHECK}" == "1" ] ; then
-    exit 0
-fi
```
yvaucher: Why removing this? It will slow down the lint tests.

travis/test_pylint
```diff
@@ -17,7 +17,18 @@ fi
...
+pylint --rcfile=${PYLINT_CONFIG_DIR}/travis_run_pylint.cfg ${CLASS_CASE} ${MODULES_TO_TEST}
 pylint_status=$?
-exit $((${pylint_status}))
+if [[ "$pylint_status" == "0" ]]; then
+  echo "Pylint OK"
+  exit 0
+else
+  echo "Pylint ERRORS"
+  exit 1
+fi
```
yvaucher: It's already in travis_run_tests that we print the result. Don't add thoses extra echo

travis/test_flake8
```diff
@@ -6,4 +6,11 @@ flake8 . --config=${FLAKE8_CONFIG_DIR}/travis_run_flake8__init__.cfg
...
 status2=$?
-exit $((${status1} || ${status2}))
+
+if [[ "${status1}${status2}" == "00" ]]; then
+  echo "Flake8 OK"
+  exit 0
+else
+  echo "Flake8 ERRORS"
+  exit 1
+fi
```
yvaucher: It's already in travis_run_tests that we print the result. Don't add thoses extra echo

yvaucher:
@moylop260 It seems there are several changes out of scope of this PR is it intended?

travis/travis_install_nightly
dreispt: The prevents from running lint ans server tests at the same time.
If uy«you are running only lint tests, this script isn't (or shouldn't be) called at all.

travis/test_pylint
dreispt: I used them during debugging, and since I found they are nice to have in the log I kept them.
No problem in removing them, but I would like a second opinion on that.

dreispt:
@yvaucher They were added by me, and IMO are necessary to provide tests. Checking that lint raises error was not a supported test case,thus the modifications.

travis/travis_install_nightly
yvaucher: Then why not adding "and not TESTS == '1'" ?
I think it's important to have a fast feedback on lint

travis/test_pylint
yvaucher: My bad I though we were already providing a message travis_run_tests

Indeed when it's ok we print nothing

travis/travis_install_nightly
dreispt: I see what you mean now. My mind was on another file.
Maybe we could:

    if [ "${TESTS}" == "0" ] ; then

@moylop260 can you fix this, to avoid a PR on your PR?

moylop260: @dreispt 
Yes of course.
Can you help me with the patch?

moylop260:
FYI I have created the script snake_case_to_CamelCase here:
https://github.com/OCA/maintainer-tools/pull/105

moylop260:
Superceded by https://github.com/OCA/maintainer-quality-tools/pull/225

yvaucher:
@moylop260 Wouldn't it be preferable to merge this one as a first step?

yvaucher:
Hum ok it is indeed also superseeded by https://github.com/OCA/maintainer-quality-tools/pull/228

pedrobaeza:
I think they are not overlapping, as this one is checking a new condition (CamelCase).

yvaucher:
@pedrobaeza this one is closed and superseeded by #226 which is currently being splitted in one PR per feature (including CamelCase)

pedrobaeza:
Yeah, I see, but I have talked with @moylop260 and he is making progressive PRs now.

PR #226, opened 2015-08-14
[REF] test_pylint: Migrate to python, add full test, add odoo lint and add new config file just to PR modules changed.

moylop260:
This PR add next changes to `test_pylint` script:
 
 - Migrate `test_pylint` from bash to pylint script.
  - Use symlinks to make the instance to can `import openerp.addons.OTHER_PATH_MODULE`
 - Add full pylint test cases.
 - Add feature to check new config file just for modules changed in your PR.
    - This allow add new pylint checks without change to red the main stable project and wait by a PR to fix new checks. (Many times we don't want to change pylint checks because we need to run faster to change all stables branches. With this feature It will check just the modules changed of a project by PR.
    - Enable CamelCase class named in pr config file.
 - Add new feature to check odoo guidelines with a pylint-plugin called odoo_lint
     - CO001 - Missing icon ./static/description/icon.png
     - CO002 - Documentation ./doc/index.rst is missing
     - CO003 - Missing ./README.rst file. Template here: %s
     - CO004 - Deprecated description in manifest file __openerp__.py
     - CO005 - Missing author required "%s"
     - CO006 - Missing required %s keys in manifest file __openerp__.py
     - CO007 - No UTF-8 coding found: Use `# -*- coding: utf-8 -*-` in first or second line of your file.
     - EO001 - Syntax error in manifest file __openerp__.py
     - EO002 - Syntax error in file ./README.rst
     - EO003 - Syntax error in file ./doc/index.rst
     - RO001 - Import `Warning` should be renamed as UserError `from openerp.exceptions import Warning as UserError`
     - WO001 - Missing coding comment
     - WO002 - Unexpected interpreter comment and execute permissions. Interpreter: %s Exec perm: %s
     - WO003 - Detected api.one and api.multi decorators together.
     - WO004 - Missing api.one in copy function.

`%s` is extracted from config_file.

NOTE: git hook don't changed in this PR because I want first feedback before of copy&paste git hook script.

PR #228, opened 2015-08-16
Allow expected errors to also handle linting

dreispt:
Expected error check is moved from server tests to the test controller `travis_run_tests`.
This allows to expect for types of errors other than server errors, such as lint errors.

Extracted from the PR #223

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,7 +26,7 @@ env:
   - VERSION="8.0" TESTS="1" LINT_CHECK="0"
 
   matrix:
-  - INCLUDE="broken_module" SERVER_EXPECTED_ERRORS="1"  # test errors are detected
+  - INCLUDE="broken_module" EXPECTED_ERRORS="1"  # test errors are detected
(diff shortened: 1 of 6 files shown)


yvaucher:
Thanks for the extraction :+1:

moylop260:
IMHO expected errors of lint should process it each lint script for compatibility with `git_hook` and use a particular variable by each script to major control.

To know if this work we will need to add some lint errors to MQT.

This change is indeed here: https://github.com/OCA/maintainer-quality-tools/pull/226
Maybe I will need to add @dreispt author in the commit of this functionallity and discuss this change in other one.

moylop260:
Other matter is that using the pylint from python you can get a counter of errors (Don't just exit(1 or 0))
Of this way you can to check that you have the exactly errors expected counters.
NOTE: To use python package we need to migrate pylint to python script, to check that works we need add lint tests cases.

moylop260:
This feature now was merged in next PR:
https://github.com/OCA/maintainer-quality-tools/pull/230

@dreispt 
Can you extract the fix part of `travis/travis_install_nightly` and create a new PR?

IMHO this PR should be closed.

PR #232, opened 2015-08-18
[IMP] run_pylint: Add feature to check new conventions just in modules changed.

moylop260:
Currently, we have a problem that if we add a new conventions in pylint-test-1 we will have all repos red in travis builds for stable branches and PR ones.

But currently we are working with next steps:
 1. We have created a new pr to oca/PROJECT-X branch 8.0
 2. We received feedabck from reviewers:
   3. Please, rename the class to `CamelCase`
   4. Please, add OCA author.
   5. Please, change your comment `encondig` by `coding`
   6. Other ones.

If we can automate it, Why not automate it?
Save time of reviewer, save time of developers just in new changes.

Then we need a middle step to add new conventions without have all repos red.
We can change the modules progressively of new conventions.

This feature run with original
`pylint pylint.cfg all_modules`
plus run other one with new conventions in just modified modules
`pylint pylint_PR.cfg modified_modules`

New conventions by example

 - CamelCase

We don't need wait to fix all repositories in all base branches to add this new convention.
We can add new conventions and check it just in new changes.
After a consider time we can move new conventions to old conventions.

travis_feature

Notice in `TRAVIS_PULL_REQUEST='true'` case the erros are old_conventions + new_conventions.

The old conventions are active in all modules in PR cases.

Why `TRAVIS_PULL_REQUEST="true"` and `TRAVIS_BRANCH`?
We don't have control on travis app :(
TRAVIS_PULL_REQUEST is a [environment variable of travis](http://docs.travis-ci.com/user/environment-variables/#Convenience-Variables)
environment variables - travis ci 2015-08-21 09-41-33
This is needed here because we need to know if [is_pull_request](https://github.com/OCA/maintainer-quality-tools/blob/87f2b269667c02af210493e929eb9524eb455907/travis/travis_run_tests#L57-L58) (Currently we have this code in stable branch) to verify new conventions just in PR cases to detect a diff with base branch `TRAVIS_BRANCH`.
environment variables - travis ci 2015-08-21 09-43-57

Now we can get git diff of `TRAVIS_BUILD_DIR`
environment variables - travis ci 2015-08-21 09-45-24

With next command if is PR:
```python
  git --git_dir=${TRAVIS_BUILD_DIR}/.git diff-index origin/${TRAVIS_BRANCH}
```
Now, we have all items changed in a PR...

PR #233, opened 2015-08-18
[REF] pylint: Enable check to validate CamelCase class name for odoo version >= 8.0

moylop260:
> Progressive PR from https://github.com/OCA/maintainer-quality-tools/pull/232
To review the real diff of this PR see last commit.

This PR enable check to validate CamelCase class name just for modules changed and just for modules of odoo version >= 8.0

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,7 +32,8 @@ env:
   - VERSION="7.0" INCLUDE="test_module,second_module" ODOO_REPO="OCA/OCB"  # ODOO_REPO usage example
   - VERSION="6.1" INCLUDE="test_module,second_module"
   - LINT_CHECK="1" TESTS="0" PYLINT_EXPECTED_ERRORS="17" TRAVIS_PULL_REQUEST="false" # Use main pylint config file
-  - LINT_CHECK="1" TESTS="0" PYLINT_EXPECTED_ERRORS="18" TRAVIS_PULL_REQUEST="true"  # Use PR pylint config file
+  - VERSION="master" LINT_CHECK="1" TESTS="0" PYLINT_EXPECTED_ERRORS="18" TRAVIS_PULL_REQUEST="true"  # Use PR pylint config file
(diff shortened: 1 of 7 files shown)


pedrobaeza:
You don't need to depends on https://github.com/vauxoo-dev/maintainer-quality-tools/commit/ed542ce2d0d138b333dad5d57dd30473291064c3 for this one, doesn't it?

moylop260:
@pedrobaeza 
I need it because this is a new conventions

nhomar:
2015-08-18 12:58 GMT-05:00 moylop260 <notifications@github.com>:

> I need it because this is a new conventions


@pedrobaeza.

1. Have the script does not mean we will run it all at once without take
care.
2. You need to decide, if we will have the rule we ned automate it, it is
more risky continue expending this job manually to all collaborators as I
exposed before.
3. We already test them in our repositories and we passed some iterations
already.

Then block this PR just becuase a non fundamented fear is not correct.

- It has unit tests to vaildate regresions.
- If I have a repository which call a class with wrong names will be red
BEFORE the PR is excecuted, and ALL the manual checks that we are doing
will be done automatically....
- The miogration is manual we are only giving a tool, not automating the
conversion itself...

What is exactly the situations where you think it can fails in order to
include such unit tests and considere in the code?

BUt a Generic "too dangerous" is not a good and productive incentive
here.... con you please be more precise?





-- 
--------------------
Saludos Cordiales

[image: --]
Nhomar Hernandez

pedrobaeza:
OK, I understand. Let's follow this path then.

dreispt:
This depends on #232 and most of the code is already there. instead of "Needs Review" It would be better marked as "WIP" (though it would be nice t have a tag for "Wainting on Dependencies"...).

moylop260:
I [promised](https://github.com/OCA/maintainer-quality-tools/pull/232#issuecomment-137435807) to @lepistone  add message clearer.
This PR check CamelCase but just show `invalid-name`

Now, with next PR https://github.com/OCA/maintainer-quality-tools/pull/234 we have next message:
`Use CamelCase "%s" in class name "%s". You can use oca-autopep8 of https://github.com/OCA/maintainer-tools to auto fix it.`

Then I close without merge this one.

PR #236, opened 2015-08-20
Mqt flexible commands to use for exemple buildout anybox odoo receipe

yvaucher:
Allow to add env var CMD_ODOO_INSTALL and CMD_ODOO_TEST to for exemple use a local buildout.cfg file creating an openerp instance using https://github.com/anybox/anybox.recipe.odoo/.

in .travis.yml you need to use installer specific to buildout:

     ...
     env:
       global:
       - CMD_ODOO_INSTALL="bin/start_openerp -d {database} --stop-after-init --init={{0}}"
       - CMD_ODOO_TEST="bin/runtests -d {database} --init={{0}}"
       - VERSION="8.0" TESTS="0" LINT_CHECK="0"
       - MODULES_DIR=${TRAVIS_BUILD_DIR}/some-dir/my-addons
    
      matrix:
      - LINT_CHECK="1"
      - SCENARIO="1"
      - TESTS="1"
    ...
    install:
      - git clone --depth=1 https://github.com/yvaucher/maintainer-quality-tools.git -b mqt-buildout ${HOME}/maintainer-quality-tools
      - export PATH=${HOME}/maintainer-quality-tools/travis:${PATH}
      - a_script_to_install_buildout

    script:
      - travis_run_tests

For this exemple your `buildout.cfg` contains or extends a config file containing:

    with_devtools = true
    test_script_name = runtests

Thus on time of execution of travis_run_tests `bin/start_openerp` and `bin/runtests` will be called.

EDIT: changed from buildout specific to more flexibility on commands


yvaucher:
FYI @lmignon

travis/test_server.py
```diff
@@ -191,18 +192,77 @@ def setup_server(db, odoo_unittest, tested_addons, server_path,
...
-                "--init", ','.join(preinstall_modules),
-                ] + install_options
+    if os.environ.get("BUILDOUT"):
+        cmd_odoo = [
+            "bin/start_openerp",
+        ]
+    else:
+        cmd_odoo = [
+            "%s/openerp-server" % server_path,
+            "--addons_path", addons_path,
```
lmignon: s/--addons_path/--addons-path

lmignon:
@yvaucher Travis is red due to a typo... (https://github.com/OCA/maintainer-quality-tools/pull/236#discussion_r37541352) 
It's nice to be able to use MQT with anybox.recipe.odoo!

yvaucher:
@lmignon Thanks fixed

lmignon:
@yvaucher Can you complete the sample file https://github.com/OCA/maintainer-quality-tools/blob/master/sample_files/.travis.yml or may be create a new one `.travis.XX.yml`. You should also  add a new test repo in https://github.com/OCA/maintainer-quality-tools/tree/master/tests/test_repo to stress this new kind of build.

lepistone:
Thanks for your work @yvaucher. I'm :-1: to this. One thing I'm not happy about in MQT growing into full-fledged python programs with logic, method, command line options, and so on, is that those are still custom scripts very specific to testing OCA modules on travis. They are coupled to travis, to odoo, and to the way OCA modules work. That means that they can't be used directly locally, of with different setups, or anything else.

IMHO this PR shows the problem: if we want to use the scripts out of the specific scenario, like for buildout here, we need 135 new lines of code.

What I would personally prefer would be to try to remove the parts specific to travis from the scripts. Do you think it would be feasible?

Thanks for your time!

yvaucher:
@lepistone Here I'll try to removed  buildout specifics but make it more flexible with env vars.

lepistone:
Hei, thanks @yvaucher! I lift my :-1: .

yvaucher:
@lepistone Refactor done

dreispt:
I haven't been converted to buildout yet, but I look forward to easily reproduce TravisCI builds locally.
Since I don't know a lot about buildout, please help me to understand this:

- are the  `CMD_ODOO_INSTALL` and `CMD_ODOO_TEST` variables something specific for Buildout use?
- can't a buildout recipe instead set values for the expected TravisCI env variables?

yvaucher:
@dreispt buildout includes a script that wrap openerp-server to define sys path with buildout eggs it also use the `<buildout_dir>/etc/openerp.cfg` file to set all parameters.
There is one scipt to launch the server: `bin/start_openerp`
And a second one to launch tests: `bin/<whatever you called it in buildout.cfg>

You could use `bin/start_openerp` for both I just wanted to let the choice. That's why if CMD_ODOO_INSTALL is not set, it will install with CMD_ODOO_TEST

Ok I guess you can set env var in buildout using
https://pypi.python.org/pypi/collective.recipe.environment/0.2.0

So you have the choice of making a specific buildout.cfg for travis setting env var or doing it in the .travis.yml. It even seems a better option to have a cleaner .travis.yml

I hope it answers your questions

guewen:
@yvaucher could you rebase?

yvaucher:
@guewen rebased

travis/test_server.py
```diff
@@ -225,20 +305,22 @@ def main(argv=None):
...
         else:
             test_loglevel = 'info'
             test_loghandler = 'openerp.tools.yaml_import:DEBUG'
     odoo_full = os.environ.get("ODOO_REPO", "odoo/odoo")
-    server_path = get_server_path(odoo_full, odoo_version, travis_home)
-    addons_path = get_addons_path(travis_dependencies_dir,
-                                  travis_build_dir,
-                                  server_path)
+    if not server_path:
+        server_path = get_server_path(odoo_full, odoo_version, travis_home)
```
dreispt: Do you dislike the compact form: ``server_path = server_path or get_server_path(...)`` ?

dreispt:
Can you check the TravisCI error? https://travis-ci.org/OCA/maintainer-quality-tools/jobs/90799804#L303

travis/test_server.py
yvaucher: Yep, I prefer it that way.
And it would make it 2 lines anyway.

travis/test_server.py
```diff
@@ -260,51 +342,36 @@ def main(argv=None):
...
             # Run test command; unbuffer keeps output colors
-            command_call = ["unbuffer"] + command
-            print(' '.join(command_call))
+            command_call = 'unbuffer {0}'.format(command)
+            print(command_call)
             pipe = subprocess.Popen(command_call,
                                     stderr=subprocess.STDOUT,
-                                    stdout=subprocess.PIPE)
+                                    stdout=subprocess.PIPE,
+                                    shell=True)
```
moylop260: Is really needed `shell=True`?

pedrobaeza:
Closing also this one. If any interest, you can reopen again.

PR #238, opened 2015-08-25
[REF] pylint: Change number of message by name of message and add multiline

moylop260:
* Numbers2name
 Many times we need to check the pylint official documentation to get pylint-number information.
 But if we use the message-short-name is easiest identify it.
 By example:
  - `duplicated-key` is more explicit that `W0109`

* Multi-lines
 - To reduce conflicts of use same line if add new or remove old checks.
 - To identify a PR diff easiest.

### DIFF
--- a/travis/cfg/travis_run_pylint.cfg
+++ b/travis/cfg/travis_run_pylint.cfg
@@ -6,6 +6,24 @@ cache-size=500
 
 [MESSAGES CONTROL]
 disable=all
+
+# Enable message and code:
+#   anomalous-backslash-in-string - W1401
+#   assignment-from-none - W1111
+#   dangerous-default-value - W0102
+#   deprecated-module - W0402
+#   duplicate-key - W0109
+#   file-ignored - I0013
+#   pointless-statement - W0104
+#   pointless-string-statement - W0105
+#   print-statement - E1601
+#   redundant-keyword-arg - E1124
+#   reimported - W0404
+#   relative-import - W0403
+#   return-in-init - E0101
+#   too-few-format-args - E1306
+#   unreachable - W0101
+
 enable=anomalous-backslash-in-string,
     assignment-from-none,
     dangerous-default-value,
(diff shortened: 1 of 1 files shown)


pedrobaeza:
Is there possibility to let as comment aside each error number?

moylop260:
@pedrobaeza 
Good point
Let me try it

moylop260:
@pedrobaeza 
Multilines config file don't support comment but I add fixed of other way.
Can you give me your opinion, please?

pedrobaeza:
OK, that can serve. Why not ordering by number instead of by name?

moylop260:
@pedrobaeza 
Added comment:
https://github.com/OCA/maintainer-quality-tools/pull/238#issue-103085045

travis/cfg/travis_run_pylint.cfg
```diff
@@ -6,7 +6,39 @@ cache-size=500
...
+    file-ignored,
+    pointless-statement,
+    pointless-string-statement,
+    print-statement,
+    redundant-keyword-arg,
+    reimported,
+    relative-import,
+    return-in-init,
+    too-few-format-args,
+    unreachable,
```
dreispt: Is it allowed to add a comments at the end of each line? 
If so I suggest to add the message code at the end of each line and remove the comment block above.

moylop260: I tested with next cases:

```conf
enable=anomalous-backslash-in-string,  # comment here
    assignment-from-none,
```

```conf
enable=anomalous-backslash-in-string,
    # comment here
    assignment-from-none,
```
And don't work.

dreispt: Pity. No other option than leaving as it is, then. Thanks for trying.

dreispt:
:+1:

moylop260:
@dreispt 
Thank you!

@pedrobaeza 
Do you have another comment?

pedrobaeza:
Nop, let's wait for another review or 5 days

moylop260:
@pedrobaeza 
Ok, I got it
Thank you

nhomar:
for me a :+1:

moylop260:
Three approvals and pr of 6 days ago... then merging.

PR #239, opened 2015-08-30
[ADD] gen_addons_table script

sbidoul:
A script that replaces a placeholder in the repository-level README.md with two tables showing the available and unported modules along with their version number and a summary (or name if summary is absent). The placeholder is preserved after running the script.

The placeholder in README.md must be like this (```...``` can be anything):
```
[//]: # (addons)
...
[//]: # (end addons)
```

The (weird) syntax for markdown comments comes from http://stackoverflow.com/questions/4823468/store-comments-in-markdown-syntax.

TODO
* [ ] insert the placeholder in all repos README.md (I volonteer to do that)
* [ ] automate the regeneration (in travis_after_success or a bot, help welcome)

### DIFF
--- a/travis/gen_addons_table.py
+++ b/travis/gen_addons_table.py
@@ -1,0 +1,99 @@
+#!/usr/env/bin/python
+from __future__ import print_function
+import ast
+import os
+import re
+
+
+MARKERS = r'(\[//\]: # \(addons\))|(\[//\]: # \(end addons\))'
+
+
+class UserError(Exception):
+    def __init__(self, msg):
+        self.msg = msg
+
+
+def sanitize_cell(s):
+    if not s:
+        return ''
+    s = ' '.join(s.split())
+    return s
+
+
+def render_markdown_table(header, rows):
+    table = []
+    rows = [header, ['---'] * len(header)] + rows
+    for row in rows:
+        table.append(' | '.join(row))
+    return '\n'.join(table)
+
+
+def replace_in_readme(readme_path, header, rows_available, rows_unported):
+    readme = open(readme_path).read()
+    parts = re.split(MARKERS, readme, flags=re.MULTILINE)
+    if len(parts) != 7:
+        raise UserError('Addons markers not found or incorrect in %s' %
+                        readme_path)
+    addons = []
+    if rows_available:
+        addons.extend([
+            '\n',
+            'Available addons\n',
+            '----------------\n',
+            render_markdown_table(header, rows_available),
+            '\n'
+        ])
+    if rows_unported:
+        addons.extend([
+            '\n',
+            'Unported addons\n',
+            '---------------\n',
+            render_markdown_table(header, rows_unported),
+            '\n'
+        ])
+    parts[2:5] = addons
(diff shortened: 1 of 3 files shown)


travis/gen_addons_table.py
```diff
@@ -0,0 +1,99 @@
...
+            render_markdown_table(header, rows_unported),
+            '\n'
+        ])
+    parts[2:5] = addons
+    readme = ''.join(parts)
+    open(readme_path, 'w').write(readme)
+
+
+def gen_addons_table():
+    readme_path = 'README.md'
```
pedrobaeza: I would start from a template file, because next time we run the script, the placeholder in README will be already replaced, so we cannot update it.

pedrobaeza:
The script looks good, but for the idea I had (be run on _travis_after_success_), has to be integrated with the full flow, providing:

* A way to avoid its run on pull requests. That can be worked the same way as Transifex script does: https://github.com/OCA/maintainer-quality-tools/blob/master/travis/travis_run_tests#L58.
* A way for writing back updated README on the branch as a commit. This can be tricky, as we will get a circular looping if we don't handle it.

travis/gen_addons_table.py
sbidoul: @pedrobaeza the script preserves the placeholder in the newly generated README, so you can run it several times.

pedrobaeza: Ah, cool then. It's better solution than having 2 files.

dreispt:
What about a OCA doc/master reference repo where this type of information would be collected into?
It doesn't need to be run by TravisCI, it could be done by a bot.
I'm also thinking on Sandy's repo build status dashboard.
Many other interesting dashboards and reports would be possible.

sbidoul:
Should I then move this script to OCA/maintainers-tools so it will be run by a bot instead of travis?

@gurneyalex @max3903

bealdav:
:+1:

moylop260:
Can you add a self-test?

lmignon:
:+1: May be you can add a self-test using https://github.com/OCA/maintainer-quality-tools/tree/master/tests/test_repo

sbidoul:
@moylop260 @lmignon before writing a test, I'd like to be sure of the repo that will host this script.

If we leave it here, in the travis directory, it implies it is a tool to be run by travis. That could be, but can travis push back the updated readme to the repository? I'd think not but I'm not sure.

It rather think it's a tool to be run manually or by a bot which has the right to push to the repo. In which case the script must go to OCA/maintainer-tools (like tx_pull which also pushes to the repositories).

What do you think?

moylop260:
@sbidoul 
You are right!

@all 
Feedback please to detect the correct project.

MT | MQT | Why
---- | ---- | ---
@pedrobaeza | :+1:  | :-1
@moylop260 | [ ] | [ ]
@Imignon  | [ ] | [ ]

lmignon:
@sbidoul  you are right :+1: for OCA/MT

sbidoul:
Superseded by https://github.com/OCA/maintainer-tools/pull/135.

PR #241, opened 2015-08-31
[REF] test_pylint: Show warning of empty paths but no error. Add colors to pylint global result.

moylop260:
Change from this
captura de pantalla 2015-08-31 a las 1 41 39 p m

to this
captura de pantalla 2015-08-31 a las 1 40 46 p m

And add red message in error case.

### DIFF
--- a/travis/test_pylint
+++ b/travis/test_pylint
@@ -8,6 +8,7 @@ Script to process pylint run and exit with result.
 import os
 
 import run_pylint
+import travis_helpers
 
 
 pylint_rcfile = os.path.join(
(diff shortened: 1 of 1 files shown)


pedrobaeza:
Thanks :+1:

travis/test_pylint
```diff
@@ -22,11 +23,13 @@ expected_errors = int(
     os.environ.get('PYLINT_EXPECTED_ERRORS', 0))
 
 exit_status = 0
-if count_errors != expected_errors:
-    print("pylint expected errors {expected_errors}, "
+if count_errors == -1:
```
dreispt: Use <0 instead?

dreispt:
:+1: Thanks!

travis/test_pylint
moylop260: We have:
`-1: python modules not found`
-2 should be a unknow error.

moylop260: Then, I'm adding a validation to get the case of `empty-folder` (-1)
I can't to use a global negative value to `empty-folder`
Maybe in future you add a new negative error:
`-2: Other handled error that no is empty folder`

dreispt:
Can we fast-track this one?

pedrobaeza:
Yeah, it's really needed

dreispt:
:green_heart:

Issue #245, opened 2015-09-14
False Error since last commit


legalsylvain:
Using OCA maintainer quality tools for a custom project, I have Travis errors since today.
https://travis-ci.org/grap/odoo-addons-it/jobs/80231337
openerp file generate W0104 errors today.

I suspect this line : 
https://github.com/OCA/maintainer-quality-tools/commit/f87fd225256d58ea227dca79ef4ab16443c55e21#diff-e10c7dd828886a45ae8d4384a7455e8fR3
from the last commit.

Did you faced the same problem ?

CC : @moylop260, @nhomar.
[V7 Travis Configuration](https://github.com/grap/odoo-addons-it/blob/7.0/.travis.yml)
Kind regards.

moylop260:
Are you using a custom pylint conf?
(reviewing...)

legalsylvain:
No, As said, this is my Travis config file.
https://github.com/grap/odoo-addons-it/blob/7.0/.travis.yml

moylop260:
You will need to add `VERSION="7.0"`
https://github.com/grap/odoo-addons-it/blob/a3fe9eb3553c2ae2092f3769caae940b9ff56729/.travis.yml#L13

legalsylvain:
Thanks, I'm trying your proposal.

moylop260:
You can view .travis.yml template here: https://github.com/OCA/maintainer-quality-tools/blob/master/sample_files/.travis.yml

moylop260:
Thank for reporting.
I will close this issue.
You can feel free to reopen

legalsylvain:
[It works now](https://travis-ci.org/grap/odoo-addons-it/jobs/80239992)
Thanks a lot.

moylop260:
@all 
This error is generated because, now pylint not ignored `__openerp__.py` file, but need of extra plugin with extra params from `VERSION` case.

I can fix this issue but if you don't add `VERSION` you code could be reviewed by lastest guidelines.
Maybe a message "`VERSION` now is required in lint tests" is better.

PR #246, opened 2015-09-14
[FIX] pylint: Fix VERSION environment variable empty case

moylop260:
Fix next cases: https://github.com/OCA/maintainer-quality-tools/issues/245

### DIFF
--- a/travis/test_pylint
+++ b/travis/test_pylint
@@ -51,12 +51,12 @@ def get_extra_params(odoo_version):
     :param version: String with name of version of odoo
     :return: List of extra pylint params
     '''
-    params = ["--load-plugins=pylint_oca"]
     odoo_version = odoo_version.replace('.', '')
     custom_cfg = os.path.join(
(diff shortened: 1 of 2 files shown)


travis/test_pylint
```diff
@@ -66,14 +66,17 @@ def get_extra_params(odoo_version):
     return params
 
 
+extra_params_cmd = ['--extra-params', '--load-plugins=pylint_oca']
```
pedrobaeza: Why change load-plugins here?

moylop260: Because the plugin disable check pointless-statement in manifest files.
Then if we don't get the version not should show pointless-statement message.

pedrobaeza: OK, got it

moylop260:
ping

pedrobaeza:
:+1:

dreispt:
:+1: LGTM

PR #263, opened 2015-10-04
[FIX] let travis tools ignore uninstallable modules

sbidoul:
The following commands need to be adapted to ignore uninstallable modules
* [x] pylint

Later:
* coverage (it seems `__unported__` is not excluded from coverage at the moment, so this can be left for another PR)
* code climate (we just have a sample file, so this can be left for a future improvement)

### DIFF
--- a/travis/run_pylint.py
+++ b/travis/run_pylint.py
@@ -9,6 +9,8 @@ import sys
 import click
 import pylint.lint
 
+import getaddons
+
 CLICK_DIR = click.Path(exists=True, dir_okay=True, resolve_path=True)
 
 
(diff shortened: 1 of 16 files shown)


pedrobaeza:
:+1:

Coverage would need to exclude dirs that are not installable. I think this can be done on https://github.com/OCA/maintainer-quality-tools/blob/master/travis/travis_after_tests_success#L11

sbidoul:
@pedrobaeza don't we need to update .coveragerc in travis_install_nightly?

pedrobaeza:
But this file is something static. We can:

* Modify that file before calling coveralls, inserting in the file the forbidden folders (uninstallable ones).
* Pass to python coveralls these folders as argument.

travis/run_pylint.py
```diff
@@ -9,6 +9,8 @@
 import click
 import pylint.lint
 
+import getaddons
+
```
moylop260: Get addons of pylint is in test_pylint file you could re-use same methods

pedrobaeza:
Any advance on coveralls part?

sbidoul:
@pedrobaeza I have not had time to progress on the coverage part yet. On the other hand, as far as I can tell, coverage does not exclude `__unported__` so the result will be the same if uninstallable modules are moved out of unported. So I think we can leave the coverage thing for another PR.

pedrobaeza:
Oh, I thought .coveragerc excludes __unported__, but I'm seeing that that's not true, so we can continue as you say and let that for another PR. What about the comment from @moylop260?

travis/run_pylint.py
sbidoul: @moylop260 I don't understand your comment. Here I'm just adding a filter to an existing method. Do you want me to do some refactoring here?

lmignon:
:+1:  (Code review)

pedrobaeza:
:+1: anyway

moylop260:
Im in California for odoo upgrade...
I will create a pr to fix my comment
Then merge

PR #270, opened 2015-10-13
[IMP] improve test_flake8 to ignore uninstallable modules

sbidoul:


yvaucher:
:+1:

lmignon:
:+1:

max3903:
:+1:

pedrobaeza:
Pylint is still checking unported modules: https://travis-ci.org/OCA/product-attribute/jobs/85247279

sbidoul:
Hmm. It seems product_extended has no `__openerp__.py`. That must be the issue.

pedrobaeza:
OK, I see, that could be solved then removing that module (as the vast majority of unported that will be finally removed...) or creating a manifest.

PR #288, opened 2015-11-03
Install lessc to fix 9.0 builds

dreispt:


.travis.yml
```diff
@@ -15,6 +15,7 @@ addons:
   apt:
     packages:
       - expect-dev  # provides unbuffer utility
+      - npm
```
moylop260: travis has npm installed by default

rvalyi:
aside from npm already being present :+1: LGTM

pedrobaeza:
We can only see the result applying this to any 9.0 branch. Can you make a test PR changing the source of MQT in Travis in any of the regular repos?

dreispt:
Self tests are supposed to prove that. Aren't they enough?

pedrobaeza:
But then you should add 9.0 line to travis

dreispt:
@pedrobaeza Yes, I'll do that.

dreispt:
@rvalyi I wasn't sure it was provided by default. But maybe we can keep to have the dependency explicit.

dreispt:
Fixed.

pedrobaeza:
The change in global doesn't affect, because there's no line that not specify the version explicitly. Please add one.

pedrobaeza:
Sorry, I didn't see first line. All is OK.

:+1:

moylop260:
:+1:

rvalyi:
:+1: @dreispt  I have nothing against letting it explicit (what I'm against is the crappy Odoo webstack itself remember :-p )

dreispt:
Can someone do the honors?

pedrobaeza:
Of course :smiley:

guewen:
Seems this commit breaks some things such as the build of the documentation using https://github.com/OCA/odoo-sphinx-autodoc.

See https://travis-ci.org/OCA/connector/jobs/89212615

If I understand correctly, npm creates a `node_modules` directory which is mingled with the addons-path. Maybe should we use the global option `-g` or use the `--prefix` option to move this directory elsewhere (disclaimer: I don't know very well `npm`, just googled a bit)?

pedrobaeza:
Or maybe blacklist `node_modules` directory in odoo-sphinx-autodoc?

guewen:
> Or maybe blacklist node_modules directory in odoo-sphinx-autodoc?

`odoo-sphinx-autodoc` is not limited to use in mqt with travis, so no, it is not a good idea to add coupling between it and mqt. Also this issue may affect other tools, so it is better if we can fix it directly in travis.

I see 2 possibilities: either we move the other repositories elsewhere than in $HOME (subdirectory) or we ensure that we have only valid addons-path folders in $HOME and then we move `node_modules` elsewhere. I think the first one is cleaner, do you see any problem with that?

dreispt:
I like the first option; don't see any reason not to implement it.

pedrobaeza:
+1 for the first option also

guewen:
Working on it: #289

PR #289, opened 2015-11-05
Store dependency repositories in $HOME/dependencies

guewen:
Instead of $HOME because other tools that expect to have a directory which
contains only valid addons paths will get confused if other directories are
created in $HOME (e.g. npm creates a 'node_modules' folder).

Follows discussion on #288

### DIFF
--- a/travis/test_server.py
+++ b/travis/test_server.py
@@ -118,15 +118,15 @@ def get_server_path(odoo_full, odoo_version, travis_home):
     return server_path
 
 
-def get_addons_path(travis_home, travis_build_dir, server_path):
+def get_addons_path(travis_dependencies_dir, travis_build_dir, server_path):
     """
     Calculate addons path
-    :param travis_home: Travis home directory
+    :param travis_dependencies_dir: Travis dependencies directory
     :param travis_build_dir: Travis build directory
     :param server_path: Server path
     :return: Addons path
     """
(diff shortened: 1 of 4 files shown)


guewen:
@dreispt @pedrobaeza could you review?

lmignon:
:+1: LGTM (Code review only)

travis/travis_transifex.py
```diff
@@ -36,6 +36,10 @@ def main(argv=None):
         return 1
 
     travis_home = os.environ.get("HOME", "~/")
+    travis_dependencies_dir = os.path.join(
+        os.environ.get("HOME", "~/"),
```
pedrobaeza: You can join here directly travis_home and 'dependencies'

travis/test_server.py
```diff
@@ -208,6 +208,10 @@ def main(argv=None):
     if argv is None:
         argv = sys.argv
     travis_home = os.environ.get("HOME", "~/")
+    travis_dependencies_dir = os.path.join(
+        os.environ.get("HOME", "~/"),
```
pedrobaeza: Use travis_home previously set

guewen: done

travis/travis_transifex.py
guewen: done

pedrobaeza:
:+1:

PR #294, opened 2015-11-08
Do not include inexistent addons path

### LINKED ISSUE #293: Travis fails with OCB + new directory layout
It seems that the new layout is incompatible with current OCB Travis: https://travis-ci.org/OCA/OCB/builds/89830498

guewen:
It allows a build to continue if a build did not create the
'$HOME/dependencies' folder beforehand.

Fixes #293

### DIFF
--- a/travis/getaddons.py
+++ b/travis/getaddons.py
@@ -60,6 +60,8 @@ def is_addons(path):
 
 
 def get_addons(path):
+    if not os.path.exists(path):
+        return []
     if is_addons(path):
         res = [path]
     else:


guewen:
I created a PR on OCB to check the correction: https://github.com/OCA/OCB/pull/369

guewen:
The build on https://github.com/OCA/OCB/pull/369 was red but because of a failing test. So this correction is working.

pedrobaeza:
Indeed, so I fast-track.

PR #296, opened 2015-11-17
Install node packages globally

guewen:
`npm install less less-plugin-clean-css` creates a `node_modules` folder in the addons path.
It makes builds that expect to only find addons in the addons path to fail.
Example:
https://travis-ci.org/OCA/connector/jobs/89892364

Using `npm install -g` makes the installation of node modules global so it doesn't pollute the addons path.

Tested on: https://github.com/OCA/connector/pull/146

### DIFF
--- a/travis/travis_install_nightly
+++ b/travis/travis_install_nightly
@@ -22,7 +22,7 @@ fi
 
 # Install less
 ln -s `which nodejs` $HOME/maintainer-quality-tools/travis/node
-npm install less less-plugin-clean-css
+npm install -g less less-plugin-clean-css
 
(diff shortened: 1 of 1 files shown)


pedrobaeza:
:+1:

lmignon:
:+1:

sbidoul:
:+1: 

However, tools should resist to non-packages or non-addons or addons that are not installable or otherwise have a wrong manifest in the addons path.

lepistone:
Agree with @sbidoul's point but I :+1: this fix for now. Thanks!

PR #297, opened 2015-11-21
[IMP] test_server: Enable store database (avoid delete it) and allow run a instance alive

moylop260:
- Useful and fully tested in next runbot-docker https://github.com/OCA/runbot-addons/pull/71
 - Useful to use travis2docker locally and runbot-travis2docker with OCA repositories
 - [REF] .travis.yml: Using cp instead mv to avoid docker/docker#4570

Maybe, this change will can be useful with other CI that use docker to build:
 - [codeship will be use docker](http://pages.codeship.com/webinars/getting-started-with-continuous-delivery-using-docker?utm_campaign=Weekly+Newsletters&utm_source=hs_email&utm_medium=email&utm_content=24008076&_hsenc=p2ANqtz-8m3QJNm6Hk560x3IwMJk94m9oLIiyB7lJNKuIEBpYJQXqajwDdLPQaBW3vj6kqWrZz1qndrnBbyYAlweFdiLxcX_rxlQ&_hsmi=24008076)
 - [shippable use docker](http://docs.shippable.com/)

### DIFF
--- a/travis/test_server.py
+++ b/travis/test_server.py
@@ -180,6 +180,7 @@ def setup_server(db, odoo_unittest, tested_addons, server_path,
                  addons_path, install_options, preinstall_modules=None):
     """
     Setup the base module before running the tests
+    if the database template exists then will be used.
     :param db: Template database name
     :param odoo_unittest: Boolean for unit test (travis parameter)
     :param tested_addons: List of modules that need to be installed
@@ -196,6 +197,7 @@ def setup_server(db, odoo_unittest, tested_addons, server_path,
         subprocess.check_call(["createdb", db])
     except subprocess.CalledProcessError:
         db_tmpl_created = True
+        print("Using previous openerp_template database.")
     if not db_tmpl_created:
         cmd_odoo = ["%s/openerp-server" % server_path,
                     "-d", db,
@@ -206,8 +208,6 @@ def setup_server(db, odoo_unittest, tested_addons, server_path,
                     ] + install_options
         print(" ".join(cmd_odoo))
         subprocess.check_call(cmd_odoo)
-    else:
-        print("Using previous openerp_template database.")
     return 0
 
 
@@ -314,18 +314,15 @@ def main(argv=None):
             if db_odoo_created and instance_alive:
                 # If exists database of odoo test
                 # then start server with regular command without tests params
-                command_start = commands[0][0]
                 rm_items = [
                     'coverage', 'run', '--stop-after-init',
                     '--test-enable', '--init', None,
                     '--log-handler', 'openerp.tools.yaml_import:DEBUG',
                 ]
-                for rm_item in rm_items:
-                    try:
-                        command_start.remove(rm_item)
-                    except ValueError:
-                        pass
-                command_call = command_start + ['--db-filter=^%s$' % database]
+                command_call = [item
+                                for item in commands[0][0]
+                                if item not in rm_items] + \
+                    ['--db-filter=^%s$' % database]
             else:
                 command[-1] = to_test
                 if is_runbot:
@@ -359,6 +356,7 @@ def main(argv=None):
                 all_errors.append(to_test)
                 print(fail_msg, "Found %d lines with errors" % errors)
         if not instance_alive:
+            # Don't drop the database if will be used later.
             subprocess.call(["dropdb", database])
 
     print('Module test summary')
--- a/travis/test_server.py
+++ b/travis/test_server.py
@@ -192,23 +192,20 @@ def setup_server(db, odoo_unittest, tested_addons, server_path,
     if preinstall_modules is None:
         preinstall_modules = ['base']
     print("\nCreating instance:")
-    start_dbtmpl = True
     try:
         subprocess.check_call(["createdb", db])
     except subprocess.CalledProcessError:
-        start_dbtmpl = False
         print("Using previous openerp_template database.")
-    if not start_dbtmpl:
-        return 0
-    cmd_odoo = ["%s/openerp-server" % server_path,
-                "-d", db,
-                "--log-level=warn",
-                "--stop-after-init",
-                "--addons-path", addons_path,
-                "--init", ','.join(preinstall_modules),
-                ] + install_options
-    print(" ".join(cmd_odoo))
-    subprocess.check_call(cmd_odoo)
+    else:
+        cmd_odoo = ["%s/openerp-server" % server_path,
+                    "-d", db,
+                    "--log-level=warn",
+                    "--stop-after-init",
+                    "--addons-path", addons_path,
+                    "--init", ','.join(preinstall_modules),
+                    ] + install_options
+        print(" ".join(cmd_odoo))
+        subprocess.check_call(cmd_odoo)
     return 0
 
 
--- a/travis/test_server.py
+++ b/travis/test_server.py
@@ -225,7 +225,7 @@ def main(argv=None):
     expected_errors = int(os.environ.get("SERVER_EXPECTED_ERRORS", "0"))
     odoo_version = os.environ.get("VERSION")
     instance_alive = str2bool(os.environ.get('INSTANCE_ALIVE'))
-    is_runbot = str2bool(os.environ.get('RUNBOT'))
+    unbuffer = str2bool(os.environ.get('UNBUFFER', True))
     if not odoo_version:
         # For backward compatibility, take version from parameter
         # if it's not globally set
@@ -325,12 +325,8 @@ def main(argv=None):
                     ['--db-filter=^%s$' % database]
             else:
                 command[-1] = to_test
-                if is_runbot:
-                    command_call = []
-                else:
-                    # Run test command; unbuffer keeps output colors
-                    command_call = ["unbuffer"]
-                command_call += command
+                # Run test command; unbuffer keeps output colors
+                command_call = (["unbuffer"] if unbuffer else []) + command
             print(' '.join(command_call))
             pipe = subprocess.Popen(command_call,
                                     stderr=subprocess.STDOUT,
--- a/travis/test_server.py
+++ b/travis/test_server.py
@@ -192,22 +192,27 @@ def setup_server(db, odoo_unittest, tested_addons, server_path,
     if preinstall_modules is None:
         preinstall_modules = ['base']
     print("\nCreating instance:")
-    subprocess.check_call(["createdb", db])
-    cmd_odoo = ["%s/openerp-server" % server_path,
-                "-d", db,
-                "--log-level=warn",
-                "--stop-after-init",
-                "--addons-path", addons_path,
-                "--init", ','.join(preinstall_modules),
-                ] + install_options
-    print(" ".join(cmd_odoo))
-    subprocess.check_call(cmd_odoo)
+    db_tmpl_created = False
+    try:
+        subprocess.check_call(["createdb", db])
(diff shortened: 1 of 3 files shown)


guewen:
I disagree to add coupling with another tool. This is the same remark than https://github.com/OCA/maintainer-quality-tools/pull/236. It was asked to remove the coupling with buildout and to make the changes needed to allow travis to make the build using a buildout. Now locally we have a script which sets the environment variables so the build can be done with buildout.

Also, it seems very specifics to your own dockerfile. Would it be possible to remove the parts specific to docker but add generic entry points (and/or reuse the one in #236) which would allow you to run with docker?

I'm :-1: for this change in this form.

moylop260:
Hello @guewen 
This PR have 2 changes:

 1. I used the current script but without delete database (because in a docker we need it).
This is the main change, if you have database in travis-ci you won't have problem by that.
 2. Method to start postgresql (start postgresql...) all dockerfiles will need a postgresql services (not specific of this one).

Could you help me to identify where you see a specific part and how I can change?
FYI I won't use a script in progress to merge https://github.com/OCA/maintainer-quality-tools/pull/236 since 3 months ago.

moylop260:
Other important matter:
I have created a docker-runbot module based on this PR here:
https://github.com/OCA/runbot-addons/pull/71

guewen:
``docker_entrypoint`` is specific to docker.

guewen:
> FYI I won't use a script in progress to merge #236 since 3 months ago.

I don't understand

moylop260:
> ``docker_entrypoint`` is specific to docker. 

Ok, I got it
I will remove this method.
Thank you.

moylop260:
@guewen 
Done!

travis/test_server.py
```diff
@@ -191,16 +191,23 @@ def setup_server(db, odoo_unittest, tested_addons, server_path,
...
-                "--init", ','.join(preinstall_modules),
-                ] + install_options
-    print(" ".join(cmd_odoo))
-    subprocess.check_call(cmd_odoo)
+    db_tmpl_created = False
+    try:
+        subprocess.check_call(["createdb", db])
+    except subprocess.CalledProcessError:
+        db_tmpl_created = True
+    if not db_tmpl_created:
```
dreispt: Use an `else:` block here instead of the `if` and get rid of the variable.

travis/test_server.py
```diff
@@ -191,16 +191,23 @@ def setup_server(db, odoo_unittest, tested_addons, server_path,
...
+                    "-d", db,
+                    "--log-level=warn",
+                    "--stop-after-init",
+                    "--addons-path", addons_path,
+                    "--init", ','.join(preinstall_modules),
+                    ] + install_options
+        print(" ".join(cmd_odoo))
+        subprocess.check_call(cmd_odoo)
+    else:
+        print("Using previous openerp_template database.")
```
dreispt: move the print to the `except` block

travis/test_server.py
```diff
@@ -296,11 +304,36 @@ def main(argv=None):
...
+        except subprocess.CalledProcessError:
+            db_odoo_created = True
         for command, check_loaded in commands:
-            command[-1] = to_test
-            # Run test command; unbuffer keeps output colors
-            command_call = ["unbuffer"] + command
+            if db_odoo_created and instance_alive:
+                # If exists database of odoo test
+                # then start server with regular command without tests params
+                command_start = commands[0][0]
```
dreispt: I think this is buggy for the unit test case.
Should be `command_start = command[0]`

travis/test_server.py
```diff
@@ -296,11 +304,36 @@ def main(argv=None):
...
+                rm_items = [
+                    'coverage', 'run', '--stop-after-init',
+                    '--test-enable', '--init', None,
+                    '--log-handler', 'openerp.tools.yaml_import:DEBUG',
+                ]
+                for rm_item in rm_items:
+                    try:
+                        command_start.remove(rm_item)
+                    except ValueError:
+                        pass
```
dreispt: Replace the 6 line for with a 1 line list comprehension: 
`command_start = [x for x in command[0] if x not in rm_items]`

Nevertheless, this feels like a hot fix and not evolution friendly.
Isn't there a way to we handle this above, when we construct the command?

dreispt:
This feature should be explained documented somewhere: in the README or at least in a code comment.

travis/test_server.py
moylop260: good catch!
I will change it

travis/test_server.py
moylop260: good catch!
Thank you
I will change it

travis/test_server.py
moylop260: Thank for your feedback...
Could you help me to understand this point?

moylop260:
@dreispt 
Thank for feedback
I used your recommendation

I add next comments:
https://github.com/OCA/maintainer-quality-tools/pull/297/files#diff-4b42f344199e5694ffee3d2a57dd80c7R183

https://github.com/OCA/maintainer-quality-tools/pull/297/files#diff-4b42f344199e5694ffee3d2a57dd80c7R315

https://github.com/OCA/maintainer-quality-tools/pull/297/files#diff-4b42f344199e5694ffee3d2a57dd80c7R359

The real documentation of how to use this features will be in [runbot-docker](https://github.com/OCA/runbot-addons/pull/71) module.

travis/test_server.py
```diff
@@ -217,6 +224,8 @@ def main(argv=None):
     install_options = os.environ.get("INSTALL_OPTIONS", "").split()
     expected_errors = int(os.environ.get("SERVER_EXPECTED_ERRORS", "0"))
     odoo_version = os.environ.get("VERSION")
+    instance_alive = str2bool(os.environ.get('INSTANCE_ALIVE'))
+    is_runbot = str2bool(os.environ.get('RUNBOT'))
```
guewen: Maybe `UNBUFFER` (active by default) instead of `RUNBOT` would be less runbot specific and allow other callers to deactivate `unbuffer` if needed?

guewen:
> @guewen
> Done!

Thanks for the change

travis/test_server.py
guewen: > Nevertheless, this feels like a hot fix and not evolution friendly.
> Isn't there a way to we handle this above, when we construct the command?

I agree with that

travis/test_server.py
```diff
@@ -296,11 +304,33 @@ def main(argv=None):
...
+                    '--log-handler', 'openerp.tools.yaml_import:DEBUG',
+                ]
+                command_call = [item
+                                for item in commands[0][0]
+                                if item not in rm_items] + \
+                    ['--db-filter=^%s$' % database]
+            else:
+                command[-1] = to_test
+                if is_runbot:
+                    command_call = []
```
dreispt: Why with "is_runbot" you remove unbuffer?

travis/test_server.py
```diff
@@ -296,11 +304,33 @@ def main(argv=None):
...
+        try:
+            db_odoo_created = subprocess.call(
+                ["createdb", "-T", dbtemplate, database])
+        except subprocess.CalledProcessError:
+            db_odoo_created = True
         for command, check_loaded in commands:
-            command[-1] = to_test
-            # Run test command; unbuffer keeps output colors
-            command_call = ["unbuffer"] + command
+            if db_odoo_created and instance_alive:
```
dreispt: We need to add on .travis.yml a test run that runs the code in this if block.

moylop260:
FYI I found other CI that use docker to build:
 - [codeship will be use docker](http://pages.codeship.com/webinars/getting-started-with-continuous-delivery-using-docker?utm_campaign=Weekly+Newsletters&utm_source=hs_email&utm_medium=email&utm_content=24008076&_hsenc=p2ANqtz-8m3QJNm6Hk560x3IwMJk94m9oLIiyB7lJNKuIEBpYJQXqajwDdLPQaBW3vj6kqWrZz1qndrnBbyYAlweFdiLxcX_rxlQ&_hsmi=24008076)
 - [shippable use docker](http://docs.shippable.com/)

travis/test_server.py
moylop260: The file of runbot is a text plain in a browser.
It when I used with `unbuffer` I saw next logfile:
`\033[1;32mINFO\033[0;m openerp_test`
But runbot use [regex](https://github.com/odoo/odoo-extra/blob/1e2d4bfdfe0e66e0a000e10edd8bf68b756eff46/runbot/runbot.py#L43-L44) without color format

travis/test_server.py
moylop260: You are right!
This section use a instance alive (a odoo server running eternally), then I don't knew how to test it easily.
We can think in a new way of test it.

By now we are testing this section in next [runbot-addons#71](https://github.com/OCA/runbot-addons/pull/71/files#diff-8cc021a18c100ed3cee5c1d675c8e723R91)
With next build: https://travis-ci.org/OCA/runbot-addons/jobs/92647163

It the first module of runbot with tests (The bad new is that don't increase the coverage of MQT) but you can give me a change of think in a way to test it here.

travis/test_server.py
moylop260: You are right!
I will change it

moylop260: @guewen 
Fixed

moylop260:
@guewen @dreispt 
Thanks for feedback.
I replied/fixed your comments.

moylop260:
@all ping

travis/test_server.py
dreispt: Can we drop in a `createdb` command somewhere to make that part of the code run?

moylop260: I don't understand you
Could you help me to explain this case?

moylop260:
FYI tests added

travis/test_server.py
moylop260: @dreispt 
Tests added

travis/test_server.py
```diff
@@ -191,7 +192,14 @@ def setup_server(db, odoo_unittest, tested_addons, server_path,
...
         preinstall_modules = ['base']
     print("\nCreating instance:")
-    subprocess.check_call(["createdb", db])
+    start_dbtmpl = True
+    try:
+        subprocess.check_call(["createdb", db])
+    except subprocess.CalledProcessError:
+        start_dbtmpl = False
+        print("Using previous openerp_template database.")
+    if not start_dbtmpl:
```
dreispt: Python has a `try ... except ... else ...` compound command: https://docs.python.org/2/reference/compound_stmts.html#try
If no exception is fired in the try block, execution will continue in the else block.
So, you don't need `start_dbtmpl`, and just replace this if with the `else:` part of the try command.

moylop260: @dreispt 
Thank for clarification
I got it
I will change it

moylop260: Fixed!

moylop260: Daniel,
Your point is very good!
Could you added to contributing?

dreispt: Its a Python pattern, its not specific to Odoo.

dreispt:
:+1:

oscarolar:
:+1:

moylop260:
@guewen 
Do you want lift your -1?

moylop260:
Can be merged?

guewen:
Yes I lift my -1. Thanks for the changes

PR #298, opened 2015-11-24
[IMP] Add images to required keys of manifest file

max3903:
Check if the module comes with screenshots to be displayed on Odoo Apps
Example: https://www.odoo.com/apps/modules/8.0/crm_voip/

### DIFF
--- a/travis/cfg/travis_run_pylint_pr.cfg
+++ b/travis/cfg/travis_run_pylint_pr.cfg
@@ -11,7 +11,7 @@ cache-size=500
 [ODOOLINT]
 readme_template_url="https://github.com/OCA/maintainer-tools/blob/master/template/module/README.rst"
 manifest_required_author="Odoo Community Association (OCA)"
-manifest_required_keys=license
+manifest_required_keys=license,images
 manifest_deprecated_keys=description,active
 
 [MESSAGES CONTROL]


moylop260:
:+1:

sebastienbeau:
:+1:

guewen:
An image is not applicable for all addons, what are we supposed to show for technical/base addons?

rvalyi:
@max3903 :+1: 

Just a side note: Long ago, I once made an extension of ooor able to take a screenshot given the xml_id of the view; I started put that in our Gollum wiki for the trainings... In a future we could imagine that the Travis run could take the screenshot and put it for doc right during the tests. I'm not sure Travis can do it. But I'm sure something like CircleCI (another CI hookable into Github) can do it. At least it won't me more complicated than something like http://gliderlabs.com/pagebuilder/latest/

Even pagebuilder itself I would like to use it for the OCA modules...

But well, let's go with manual screenshot first and then we can imagine that may be...

rvalyi:
@guewen kitten or other subversive images then.

guewen:
> @guewen kitten or other subversive images then.

All right! Easy enough :-D

max3903:
@guewen Any example in particular ? Maybe the configuration screen then.

yvaucher:
For example of module for which an image shouldn't be required:

All modules starting with `base_` which are there to provide an interface for other modules or are only technical changes like patches. (ok it might be easy to find something for `base_continent`)

Few examples:
`base_delivery_carrier_label` https://github.com/OCA/carrier-delivery/tree/9.0/base_delivery_carrier_label
`base_concurrency` https://github.com/OCA/server-tools/tree/9.0/base_concurrency
`web_context_tunnel` https://github.com/OCA/server-tools/tree/9.0/web_context_tunnel


@rvalyi I let you find the right kitten :smirk_cat: for `web_context_tunnel` if you still need it in 9.0 (I'm afraid not all onchange are gone but that's another story)

dreispt:
As a Warning only, please.

moylop260:
@dreispt 
This checks is a warning only.

warning


https://travis-ci.org/OCA/maintainer-quality-tools/jobs/93032168#L324

guewen:
> @rvalyi I let you find the right kitten :smirk_cat: for web_context_tunnel if you still need it in 9.0 (I'm afraid not all onchange are gone but that's another story)
kitten-tunnel

yvaucher:
@guewen :+1:

rvalyi:
@guewen :+1: lol

StefanRijnhart:
Can we revert this? The images key is not for screenshots. It is for the splash screen of the module on the apps store. Just putting in any screenshot gives an ugly result. The default coloured box with the module icon looks a lot better in many cases.

pedrobaeza:
@StefanRijnhart, as I have said in the other PR, this is only a warning, it's not mandatory to put the images key.

StefanRijnhart:
Well, I still think it will encourage people to put a random screenshot there, making it look worse on the apps store.

pedrobaeza:
For now people is not attending too much this I'm afraid, but we can improve the warning text or put a contributing section about that (and a link to it in the warning).

lepistone:
Sorry I came late to the party, but I :-1: the `images` check (so I :+1: @StefanRijnhart "can we revert this").

For me "warning" means something is wrong, but for some reason it doesn't crash the program (or the build or whatever). It is still something you want to get rid of. So in general I think we shouldn't add warnings unless we think they should be enforced one day.

In that case I have the impression it is not always relevant, especially if it is a random screenshot (much less important than documentation, for example).

max3903:
@lepistone What about using "info" instead of "warning"?

moylop260:
FYI I created the PR to continue this discuss:
https://github.com/OCA/maintainer-quality-tools/pull/315

PR #300, opened 2015-11-28
[REF] clone_oca_dependencies: Add feature to upgrade git version of repositories dependencie

moylop260:
- If you run 2 times the `clone_oca_dependencies` script you will get:
  - First time: `git clone ...`
  - Second time: `git pull ...` 
 - Update repositories dependencies previously downloaded.
 - Useful when you use MQT locally to upgrade repositories

### DIFF
--- a/travis/clone_oca_dependencies
+++ b/travis/clone_oca_dependencies
@@ -58,8 +58,11 @@ def git_checkout(deps_checkout_dir, reponame, url, branch):
     checkout_dir = osp.join(deps_checkout_dir, reponame)
     if not osp.isdir(checkout_dir):
         command = ['git', 'clone', '-q', url, '-b', branch, checkout_dir]
-        _logger.info('Calling %s', ' '.join(command))
-        subprocess.check_call(command)
+    else:
+        command = ['git', '--git-dir=' + os.path.join(checkout_dir, '.git'),
+                   '--work-tree=' + checkout_dir, 'pull', url, branch]
+    _logger.info('Calling %s', ' '.join(command))
+    subprocess.check_call(command)
     return checkout_dir
 
 
--- a/travis/clone_oca_dependencies
+++ b/travis/clone_oca_dependencies
@@ -60,7 +60,8 @@ def git_checkout(deps_checkout_dir, reponame, url, branch):
         command = ['git', 'clone', '-q', url, '-b', branch, checkout_dir]
     else:
         command = ['git', '--git-dir=' + os.path.join(checkout_dir, '.git'),
-                   '--work-tree=' + checkout_dir, 'pull', url, branch]
+                   '--work-tree=' + checkout_dir, 'pull', '--ff-only',
+                   url, branch]
     _logger.info('Calling %s', ' '.join(command))
     subprocess.check_call(command)
     return checkout_dir


travis/clone_oca_dependencies
```diff
@@ -60,6 +60,11 @@ def git_checkout(deps_checkout_dir, reponame, url, branch):
         command = ['git', 'clone', '-q', url, '-b', branch, checkout_dir]
         _logger.info('Calling %s', ' '.join(command))
         subprocess.check_call(command)
+    else:
+        command = ['git', '--git-dir=' + os.path.join(checkout_dir, '.git'),
```
dreispt: Are you sure about join `.git`? AFAIK it should be just the repo's root directory...

moylop260: desctop screenshot

moylop260: desctop screenshot2

dreispt: ok

dreispt:
:+1: LGTM

sylvain-garancher:
The `_logger.info` and `subprocess.check_call` lines are the same, maybe you can get them out of the `if` statement ?

travis/clone_oca_dependencies
sylvain-garancher: When using the `--git-dir` argument, you should use `--work-tree`, to define the working directory of your git repository.

Try calling `git --git-dir=/some/path/.git` from outside your git repository. The changes will be applied in your current working directory, instead of the git repository.

moylop260:
@sylvain-garancher 
Good point, thank you.
Fixed!

travis/clone_oca_dependencies
moylop260: I'm not sure...
We are using this sentence since along time ago in runbot:
https://github.com/odoo/odoo-extra/blob/1e2d4bfdfe0e66e0a000e10edd8bf68b756eff46/runbot/runbot.py#L224

2015-12-01 9:35 GMT-05:00 Sylvain Garancher <notifications@github.com>:

> In travis/clone_oca_dependencies
> <https://github.com/OCA/maintainer-quality-tools/pull/300#discussion_r46284620>
> :
>
> > @@ -60,6 +60,11 @@ def git_checkout(deps_checkout_dir, reponame, url, branch):
> >          command = ['git', 'clone', '-q', url, '-b', branch, checkout_dir]
> >          _logger.info('Calling %s', ' '.join(command))
> >          subprocess.check_call(command)
> > +    else:
> > +        command = ['git', '--git-dir=' + os.path.join(checkout_dir, '.git'),
>
> When using the --git-dir argument, you should use --work-tree, to define
> the working directory of your git repository.
>
> Try calling git --git-dir=/some/path/.git from outside your git
> repository. The changes will be applied in your current working directory,
> instead of the git repository.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/OCA/maintainer-quality-tools/pull/300/files#r46284620>
> .
>



-- 
Moisés López Calderón
Vauxoo - OpenERP's Gold Partner
Mobile: (+521) 477-752-22-30
Office: (+52) 477-773-33-46
web: http://www.vauxoo.com
twitter: @vauxoo
           @moylop260
hangout: moylop260@vauxoo.com

sylvain-garancher: For git commands used in runbot, like `gc`, `log` or `fetch`, which don't use a working tree, this works fine.

But `pull` uses a working tree, which is the current working directory by default.

lepistone:
Thanks for the patch. I am :-1: to this, because making this solid enough to be useful needs to handle many, many cases. For example:

- what if the version needed is older than what is checked out at the moment?
- what if some working directories were dirty?
- what if the user has committed something locally?
- what if the user changed the remote, for example, to his own fork to make a pull request?
- how to we tell the user that for some reason something went wrong?
- what if the last pull had conflicts, and what if we run it again?

A few of these cases can be somewhat avoided by adding the `--ff-only` option, but I feel that simple wrappers to `git` tend to create more trouble than they solve.

At the very least, there should be some documentation saying in what case we can expect the update to work, and what will happen then.

moylop260:
@lepistone 
Thank you for you feedback.
The main focus of this PR is just avoid run `git pull` in each path of `oca_dependencies.txt` for the people that use it in ideal cases.
This isn't a full `git` wrapper (for those you can use `git` manually and you can avoid use this script).

NOTE: `oca_dependencies` aren't your work directory, your work directory is `travis/build/OCA/project` (to create changes, commits, push, change of remotes...).

With this info,
Could you give me that points I should consider fix to merge this one?

lepistone:
@moylop260 if we merge this, I'd propose what I said in the other message:
- `--ff-only`
- tell clearly what this does, including the update feature, and when it is supposed to work. With this patch, the description at the beginning of the file doesn't correspond anymore to the behavior.

Thanks!

moylop260:
@lepistone 
I got it
Thank you

moylop260:
@lepistone 
FYI changes of your feedback applied
 - Change name of pr and add more info in description 
 - Add `--ff-only` parameter in `git pull`

Thank for feedback

moylop260:
ping

moylop260:
@pedrobaeza @osvalr @max3903 @dreispt your feedback is welcome

dreispt:
I see this useful for local tests, and keep my :+1:

lepistone:
sorry for not answering @moylop260, I missed the notification. Thanks for the change! I lift my 👎 .

The fact that I prefer to not use scripts like this (because they get me confused) shouldn't stop others that find them useful... So no problem from me!

moylop260:
@lepistone thanks
Then could we merge it?

osvalr:
:+1: for me

PR #305, opened 2015-12-30
[FIX] fix regex matching for error messages

tremlin:
The current matching of error messages seems to have two problems:
1. Partial string matching must use the `search` method of `re` (instead of `match`)
2. The lambdas in the loop don't work correctly because of late binding (workaround suggested by http://docs.python-guide.org/en/latest/writing/gotchas/#late-binding-closures)

This pull requests attempts to solve the problem and adds a new test case.


tremlin:
I know that two test cases fail, but this is by design and I would appreciate some feedback on how to proceed.
The only test cases that fail are those in pull request-mode. As far as I can see, the (new) errors come from the fact that I touched the broken modules by adding model access rights. Adding access rights to all existing models and adding a new module without rights is in my opinion be the best way to organize the tests to keep them as orthogonal as possible.

OSguard:
@gurneyalex @guewen @pedrobaeza @moylop260 can we discuss this issue?

moylop260:
@OSguard 
Of course!
Let me check in a travis2docker local what is the issue with this matter

travis/test_server.py
```diff
@@ -43,10 +43,10 @@ def make_pattern_list_callable(pattern_list):
         for i in range(len(pattern_list)):
             if isinstance(pattern_list[i], basestring):
                 regex = re.compile(pattern_list[i])
-                pattern_list[i] = lambda x: regex.match(x['message'])
+                pattern_list[i] = lambda x, regex=regex: regex.search(x['message'])
```
moylop260: Fix `E501 line too long (83 > 79 characters)`

travis/test_server.py
```diff
@@ -43,10 +43,10 @@ def make_pattern_list_callable(pattern_list):
...
             elif hasattr(pattern_list[i], 'match'):
                 regex = pattern_list[i]
-                pattern_list[i] = lambda x: regex.match(x['message'])
+                pattern_list[i] = lambda x, regex=regex: regex.search(x['message'])
```
moylop260: Fix `E501 line too long (83 > 79 characters)`

moylop260:
The problem detected is that you change broke the check of flake8 `Fix `E501 line too long` in `test_server.py` file.
This check don't is very clear in the log, but you can find it here:
https://travis-ci.org/OCA/maintainer-quality-tools/jobs/99461215#L514-L515

moylop260:
FYI in my container the new version of flake8 is installed maybe next travis build will show follow checks:
```bash
flake8 . --config=${FLAKE8_CONFIG_DIR}/travis_run_flake8.cfg
./travis/test_server.py:46:80: E501 line too long (83 > 79 characters)
./travis/test_server.py:49:80: E501 line too long (83 > 79 characters)
./travis/test_server.py:175:17: W503 line break before binary operator
./travis/test_server.py:176:17: W503 line break before binary operator
status2=$?
```

screen shot 2016-02-18 at 2 34 55 pm

tremlin:
@moylop260 Thanks for your detailed explanation. I have to improve my log reading skills.

Now at least all tests are green.

moylop260:
@tremlin 
Really we have a ci with ugly log.
But work for now.

Thank for fixing

moylop260:
:+1:

pedrobaeza:
:+1:

PR #307, opened 2016-01-14
Allow osv.expression

yvaucher:


yvaucher:
Ok we are not using the last version of pylint, we can't use the fix, I'm closing this

PR #309, opened 2016-01-21
don't overwrite existing server config

gurneyalex:
This is required e.g. for OCA/server-tools: travis config generates a
server configuration file, which is required by the server_environment
module. f4d9198 would overwrite that file causing build failures.

### DIFF
--- a/travis/test_server.py
+++ b/travis/test_server.py
@@ -226,13 +226,20 @@ def run_from_env_var(env_name_startswith, environ):
 
 
 def create_server_conf(data, version):
-    '''Create default configuration file of odoo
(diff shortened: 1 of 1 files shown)


pedrobaeza:
If there are values that are duplicated from the existing config and the one written by Travis, which value is taken? The last one?

moylop260:
:+1:

lmignon:
:+1:

gurneyalex:
@pedrobaeza yes.

damdam-s:
:+1:

lepistone:
thanks :+1:

PR #311, opened 2016-02-11
[IMP] clone_oca_dependencies: pip install requirements.txt

moylop260:
- If main project and/or projects depends exists a `requirements.txt` then will be used to install.
   - `pip install -r requirements.txt`
 - Allow install depends recursively between projects.
 - Allow use a global standard to install python dependencies.


guewen:
:+1: thanks

pedrobaeza:
:+1:

PR #310, opened 2016-02-04
Add a Travis' environment variable to set a default configuration file for pylint

lorenzo-pasa:
Replace the hardcoded value for the pylint configuration file (i.e. `travis_run_pylint.cfg`) with an environment variable, making it possible to choose a different configuration (even per Travis' .yaml, if needed).

Using the "PYLINT_CONFIG_FILE" environment variable, it's possible to choose which configuration file will be used _during the regular lint check_ (e.g. when the commit triggering Travis doesn't come from a Pull Request).

This way, it's possible to dynamically set a default pylint configuration.

PR #312, opened 2016-02-11
Add env variable to set default pylint config file

nhldn:
Using the "PYLINT_CONFIG_FILE" environment variable,
it's possible to choose which config file will be used
during the regular lint check (i.e. not for Pull Request use case).


nhldn:
This Github account belongs to Neova Trading (Health) Ltd, which already has signed and provided the Entity CLA.

moylop260:
The previous PR with all `+1` is: https://github.com/OCA/maintainer-quality-tools/pull/310
But now is signed CLA.
Then merging...

PR #315, opened 2016-02-15
Revert "[IMP] Add images to required keys of manifest file"

moylop260:
Reverts OCA/maintainer-quality-tools#298

Opened discuss from @lepistone comment:
https://github.com/OCA/maintainer-quality-tools/pull/298#issuecomment-176651653

>Sorry I came late to the party, but I :-1: the `images` check (so I :+1: @StefanRijnhart "can we revert this").
For me "warning" means something is wrong, but for some reason it doesn't crash the program (or the build or whatever). It is still something you want to get rid of. So in general I think we shouldn't add warnings unless we think they should be enforced one day.
In that case I have the impression it is not always relevant, especially if it is a random screenshot (much less important than documentation, for example).

### DIFF
--- a/travis/cfg/travis_run_pylint_pr.cfg
+++ b/travis/cfg/travis_run_pylint_pr.cfg
@@ -11,7 +11,7 @@ cache-size=500
 [ODOOLINT]
 readme_template_url="https://github.com/OCA/maintainer-tools/blob/master/template/module/README.rst"
 manifest_required_author="Odoo Community Association (OCA)"
-manifest_required_keys=license,images
+manifest_required_keys=license
 manifest_deprecated_keys=description,active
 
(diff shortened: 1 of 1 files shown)


StefanRijnhart:
Appreciated! :+1:

moylop260:
@lepistone 
Your feedback is welcome.

hbrunn:
:+1:

yvaucher:
:+1:

lepistone:
As mentioned in the other PR, I prefer not to have images required, and in general I prefer to have warnings for things that really we want to get rid of one day. In other words, I think warning means "we really want that, but maybe we have no time to do that now". For images, I don't think we always want them.

Thanks! :+1:

PR #316, opened 2016-02-15
[IMP] pylint_pr.cfg: Enable check eval-used - W0123

moylop260:
@dreispt and @gurneyalex request from [twitter](https://twitter.com/reis_pt/status/699238110375366657?cn=bWVudGlvbg%3D%3D&refsrc=email)

- [x] Enable check W0123, just in modules changed:
  - screen shot 2016-02-15 at 12 24 24 pm
  - https://travis-ci.org/OCA/maintainer-quality-tools/jobs/109412780#L471

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,8 +32,8 @@ env:
   - VERSION="7.0" INCLUDE="test_module,second_module" ODOO_REPO="OCA/OCB"  # ODOO_REPO usage example
   - VERSION="6.1" INCLUDE="test_module,second_module"
   - LINT_CHECK="1" TESTS="0" PYLINT_EXPECTED_ERRORS="15" TRAVIS_PULL_REQUEST="false" # Use main pylint config file
-  - VERSION=master LINT_CHECK="1" TESTS="0" PYLINT_EXPECTED_ERRORS="15" TRAVIS_PULL_REQUEST="true"  # Use PR pylint config file
-  - VERSION="7.0" LINT_CHECK="1" TESTS="0" PYLINT_EXPECTED_ERRORS="15" TRAVIS_PULL_REQUEST="true"  # To check pylint_conf of PR's with old api
+  - VERSION=master LINT_CHECK="1" TESTS="0" PYLINT_EXPECTED_ERRORS="16" TRAVIS_PULL_REQUEST="true"  # Use PR pylint config file
+  - VERSION="7.0" LINT_CHECK="1" TESTS="0" PYLINT_EXPECTED_ERRORS="16" TRAVIS_PULL_REQUEST="true"  # To check pylint_conf of PR's with old api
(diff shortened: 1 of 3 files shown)


max3903:
@moylop260 in the message, can you mention what to use instead? Something like `Use of eval. Use save_eval instead.`

moylop260:
@max3903 
Thanks for comment.
The message is original from `pylint` app then I don't have control for this message.

sbidoul:
:+1:

max3903:
@moylop260 ok, :+1:

pedrobaeza:
These comments from @cedk in Twitter can be interesting in this topic:

https://twitter.com/cedrickrier/status/699294548191744000

moylop260:
@pedrobaeza 
Thanks for comment.

- `eval` is major security issue, because you have full access to total environment of the program from string of eval, because use your `globals()` and `locals()`.
 - `safe_eval` is minor security issue, because you have full access to all variable that you want send it, if you send the full object is a problem of security yet, but is minor that use `eval`, here you don't have access to `globals()` or `locals()`.

Then IMHO we should avoid use `eval`.

dreispt:
@gurneyalex shown disagreement of Twitter about adding this.
In fact there are legit uses for `eval` although I believe they are rare.
Would enforcing the check and using a lint ignore marker on the legit lines be fair enough?

moylop260:
I don't know a legit use of `eval` where don't can be replaced by other one.

I wish to know them.

@gurneyalex Please, Could you help me to understand your point?

dreispt:
How about this:
- the expression to use is safe eval'd on write on the Model storing the rule
- later when the rule is used, since it's now trusted it can just be eval'd, allowing for better performance.

gurneyalex:
I won't object to this PR. Although I'm concerned that `safe_eval` may not be as safe as the name implies. It is definitely harder to exploit than the Python built-in, so let's have this check enabled.

lepistone:
:+1: in principle: if someone really needs to `eval`, I'd love an explicit ignore on the line, with a commentary telling why there's no other way. `safe_eval` is not perfect, but definitely better.

A practical thing before merging: does this check allow the common pattern `import safe_eval as eval; eval(something)` ? I don't like the pattern and personally prefer to use `safe_eval` explicitly, but I would not block for that.

Thanks!

gurneyalex:
@lepistone I don't want import safe_eval as eval as it makes reviewing more complicated (you have to check the imports at the top of the file). 

For an example of `safe_eval` behaving differently from `eval`, see https://github.com/OCA/reporting-engine/pull/45

lepistone:
We agree `import safe_eval as eval` is a bad idea. Unless we see a lot of branches becoming red, let's merge this. :+1: Thanks!

pedrobaeza:
:+1: merging then and see in future PRs what happens

dreispt:
PS: Unfortunately this a big Python outstanding issue: no support to run sandboxed code :cry:

PR #326, opened 2016-04-07
[IMP] pylint: Enable check to verify duplicate id in ir.model.access.csv files

lescobarvx:
[IMP] pylint: Enable check to verify duplicate id in ir.model.access.csv files

### DIFF
--- a/tests/test_repo/broken_module/security/ir.model.access.csv
+++ b/tests/test_repo/broken_module/security/ir.model.access.csv
@@ -1,2 +1,5 @@
 "id","name","model_id:id","group_id:id","perm_read","perm_write","perm_create","perm_unlink"
 "access_test_model","access_test_model","model_test_model","base.group_user",1,1,1,1
+"access_test_model2","access_test_model2","model_test_model","base.group_user",1,1,1,1
+"access_test_model2","access_test_model2","model_test_model","base.group_user",1,1,1,1
+"access_test_model4","access_test_model4","model_test_model","base.group_user",1,1,1,1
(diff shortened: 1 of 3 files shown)


tests/test_repo/broken_module/security/ir.model.access.csv
```diff
@@ -1,2 +1,5 @@
 "id","name","model_id:id","group_id:id","perm_read","perm_write","perm_create","perm_unlink"
 "access_test_model","access_test_model","model_test_model","base.group_user",1,1,1,1
+"access_test_model2","access_test_model2","model_test_model2","base.group_user",1,1,1,1
+"access_test_model2","access_test_model3","model_test_model3","base.group_user",1,1,1,1
+"access_test_model4","access_test_model4","model_test_model4","base.group_user",1,1,1,1
```
dreispt: I think that the model used should be `model_test_model` in all lines.

dreispt:
Don't you need to bump the expected errors counter?

tests/test_repo/broken_module/security/ir.model.access.csv
moylop260: This is a module broken... We can use a bad model name here

dreispt: That will raise an error, but for the wrong reason (invalid model), thus not performing the check on the error we want. You're not testing the duplicate ID.

moylop260: You are right!
(I don't saw the build result, my bad... fixing...)

moylop260:
Expected errors counter increased, fix csv and rebased.

dreispt:
LGTM :+1:

nhomar:
:+1:

moylop260:
Included from https://github.com/OCA/maintainer-quality-tools/commit/22a6a0bce39705942b112cd1e240ee00c1ecee9b

dreispt:
I would rather prefer that we perform regular Merge: it makes clearer what was the PR that originated the commits, and we can then refer to it's discussion for more insights. 
IM(H)O fast forward merges are good for projects with a small well defined team (or one guy), not for repos with such a large and distributed responsibility.

lasley:
They also break GPG verification from the original author & force the merger to deal with the commit note. Cool feature, but not really viable IMO

PR #337, opened 2016-06-27
[REF] Dependency cloning more extensible

dreispt:
This is a preparatory work, to next add a "satellite repo integration" feature.
The `run()` function both cloned dependency repos, and pip installed the Python requirements.txt.
Here we split these two jobs into their own functions: `clone_repos()` and `install_requirements()`.

### DIFF
--- a/travis/clone_oca_dependencies
+++ b/travis/clone_oca_dependencies
@@ -34,74 +34,80 @@ import logging
 _logger = logging.getLogger()
 
 
-def parse_depfile(depfile, owner='OCA'):
+def parse_depfile(depsfile_path, owner='OCA'):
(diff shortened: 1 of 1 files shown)


moylop260:
:+

moylop260:
👍

pedrobaeza:
Is this still relevant? If so, you need to rebase.

dreispt:
@pedrobaeza Yes it still is, but I'm following the pip-installation work to see how it matches with this.

nhomar:
@dreispt Is this valid yet? can you rebase if Yes,, you had the +1 from @moylop260 but it is not mergeabel

PR #338, opened 2016-06-28
Add automatic merge integration of proposed satellite repos

dreispt:
Contains #337

### DIFF
--- a/sample_files/oca_modules.txt
+++ b/sample_files/oca_modules.txt
@@ -1,0 +1,15 @@
+# List the OCA satellite projects to merge into this repo
+# Format <repo identifier> <repo git url> <repo branch>
+
+# Examples
+# ========
+#
+# To integrate the modules in the 'OCA/web' repository, use:
+# web
+#
+# To explicitely give the URL of a fork, and still use the version specified in
+# .travis.yml, use:
+# web https://github.com/OCA/web.git
+#
+# To provide both the URL and a branch, use:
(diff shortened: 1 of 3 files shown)


moylop260:
Sorry, I don't understand this change.
Do you have a documentation to read it or a real example?

sbidoul:
@dreispt I'm all in favor of smaller repos. But I don't understand this PR. Is it to merge the small repo in the big one automatically?

pedrobaeza:
I close this one until we have clearer how to implement this.

dreispt:
@sbidoul The point is to allow to run integration tests of modules from different repos.
Example: from OCA/project we created spin-off repos OCA/project-gtd and OCA/project-service.
But we want these repos to run installation and Python tests together.
So, for example, "OCA/project" could be told to get the modules from OCA/project-gtd and OCA/project-service and run the tests on all three on the same build.

This way, an "integration repo" is able to oversee what is going on on their satellite repos.
Does this make sense?

sbidoul:
@dreispt yes I see the need for integration testing. Thanks for the clarification.

dreispt:
@sbidoul Good. I was afraid it could be far fetched 😄  
This mechanism could also be used for an eventual OCX to be tested along with some (or all) OCA addons.

sbidoul:
@dreispt maybe we don't need a full fledged repo to do integration testing. Perhaps a simple requirements files of addons to be tested together could be sufficient.

Issue #290, opened 2015-11-05
Enable JS coverage


moylop260:
Somebody knows how to enable coverage of javascript?

dreispt:
Maybe @hbrunn can help.

lasley:
Assuming this is similar to coverage in a normal JS lib & not something insanely Odoo specific, I can likely help here once we get support for multiple languages per repo.

lasley:
Alright reviving this now that we can have multiple language support due to CodeCov.

I'm almost positive I can get us coverage reporting, but I only know how to do this with Instanbul paired with a Test Runner like Karma or Grunt. 

Looking at the Odoo source, I think our key is the fact that [Odoo also uses Grunt](https://github.com/odoo/odoo/blob/7682760dcad232829fbb5d4221f66a1ebb9d6c91/addons/web/Gruntfile.js#L3). The requirements boil down to using something like this - https://github.com/afonsof/grunt-qunit-cov

moylop260:
Excellent
Good to know
Thank you!

pedrobaeza:
Any update on this?

moylop260:
@lasley comment us that this feature require too much patches to odoo core.
(Could you help me to explain it better @lasley )

lasley:
All solutions I can find at this point involve injecting data into the [Odoo Gruntfile](https://github.com/odoo/odoo/blob/7682760dcad232829fbb5d4221f66a1ebb9d6c91/addons/web/Gruntfile.js#L3). I'm at a loss on how to do this without completely re-implementing the file, or dirty patches with SED 😦

hbrunn:
can't you wrap `module.exports` into a function that passes a fake grunt object to the original, records/passes through what Odoo's gruntfile does, and adapts afterwards the way we need it?

lasley:
Hmmmmmm I feel like I should have tried this, but I'm almost positive I haven't now that you say it. Will report back!

nhomar:
I think thi sissue can be closed, this is working already AFAICR, open it again if not.

pedrobaeza:
No, this is not working. One thing is JS linter. Another is the test coverage. It's even more important in v11 now that QUnit allows to test better JS widgets and interface.

yelizariev:
Simple solution I use myself:
https://github.com/it-projects-llc/pylint-odoo/commit/00b068817724ec756b1425356f178705200c12e7
It only checks whether there is any js coverage and makes warning otherwise

[...]

PR #342, opened 2016-07-12
[ADD] Enable codecov: Show a extra github statuses focused for just changes

moylop260:
Coveralls.io is failing too much recently.
I good example is this PR (I don't have changes affecting by monitored py files):
coveralls

Codecov adds an additional and useful github statuses to split a average of coverage just of your diff.

codecov

 - Codecov have integrate [browser plugins](https://github.com/codecov/browser-extension)

### DIFF
--- a/travis/travis_after_tests_success
+++ b/travis/travis_after_tests_success
@@ -4,8 +4,10 @@ import os
 
 from coverage.cmdline import main as coverage_main
 from coveralls import cli as coveralls_cli
+from codecov import main as codecov_main
 
 if (os.environ.get('TESTS', '1') == '1' and
         not os.environ.get('LINT_CHECK') == '1'):
     coverage_main(["report", "--show-missing"])
-    exit(coveralls_cli.main(argv=None))
+    coveralls_cli.main(argv=None)
+    codecov_main(argv=None)
--- a/travis/travis_install_nightly
+++ b/travis/travis_install_nightly
@@ -43,7 +43,7 @@ tar -xf odoo.tar.gz -C ${HOME}
 rm -f $VIRTUAL_ENV/lib/python2.7/no-global-site-packages.txt
 
 pip install --user -q -r $(dirname ${BASH_SOURCE[0]})/requirements.txt
-pip install -q QUnitSuite coveralls
+pip install -q QUnitSuite coveralls codecov
 
 # Use reference .coveragerc
 cp ${HOME}/maintainer-quality-tools/cfg/.coveragerc .


moylop260:
@OCA/core-maintainers What do you think about enable `codecov`? (See main description)

dreispt:
Sounds good to me.
This means changing the coverage badge on each repo.
Maybe this should be an opt-in feature, to be gradually enabled by maintainers, together withthe changes on the repo README.

moylop260:
@dreispt 
Thanks for reply.

> Maybe this should be an opt-in feature

I don't understand, What opt-in feature are you talking about?
Could you help me to explain it?

FYI this change don't disable coveralls, just enable codecov, then the badge of coveralls will works, we can add the new badge of codecov gradually.

To enable the repository just we need a login from a member of @OCA/core-maintainers team and sync all repos (one-by-one is not required)

dreispt:
I mean, that codecov should be explicitly enabled on the `.travis.yml`.
No point in keeping both at the same time on a repo, so when one is enabled, the other should be disabled .

moylop260:
@dreispt 
Thanks for clarify.

> No point in keeping both at the same time on a repo

I'm agree

> so when one is enabled, the other should be disabled .

We could use a environment variable to detect if use coveralls.yml or codecov.yml
but this could be more code unnecessary for a future deprecated tool.

IMHO we could live with both by a short time to test the service and generate the badge switch.
After that, we could disable coveralls from MQT.

What do you think?

dreispt:
@moylop260 No problem with that, as long as it doesn't add complexity to the technical implementation.

gurneyalex:
I could also do a mass commit to change the travis configuration files on all our projects to update the coverage tool. That's quite easy to do, and rather more efficient than a slow transition.

moylop260:
@gurneyalex 
Thanks
I'm agree with that.

By now, could we merge this one?

pedrobaeza:
From what I see, we don't need to make any change in .travis.yml, as both are executed right now. Isn't it?

moylop260:
It is!
We dont need change in .travis.yml
Just change the README project to add the badge.

lasley:
I'm a little late to this, but I just found out today that Code Climate also does coverage. Have we looked into just adding it into the existing Code Climate integrations we already have? Seems like we're stacking a lot of third parties that someone has to setup at some point.

Not blocking or anything, just throwing out ideas. Is this something we have looked into?

elicoidal:
I am OK with any tool but would ask for:

* stability (not changing tool over and over). Maybe coverall is not really working now but do we know the reasons and / or is it temporary
* One tool per concept (no need to have 3 tools for coverage, which one should be open for debate)

lasley:
Alright so I took a look deeper into the features, and I am now all for CodeCov. 👍 

Main reason IMO is that they support multiple languages in the same repo, which is something neither of the other solutions offer. This will allow for #290 to be a possibility, which is currently is not (unless we prefer JS over Python, or hack it in)

dreispt:
(OT) Looking at the checks list, Coveralls is trying to block this PR 😉

pedrobaeza:
hehe, :+1: for this

moylop260:
ping

lasley:
Hmm I see enough thumbs here, and enough time has passed too. Seems like a solid merge, yeah?

Are there any action items other than [the mass update](https://github.com/OCA/maintainer-quality-tools/pull/342#issuecomment-237159528) after merge?

pedrobaeza:
There's no need of mass commit in theory.

pedrobaeza:
Merging for applying this to see if we don't have issues like in coveralls

lasley:
Dead thread, but found this gem today that I'm sure anyone tagged on this thread might want. CodeCov has a [browser extension](http://docs.codecov.io/docs/browser-extension) that provides inline coverage in PRs. I'm in love, so I needed to share 👍

PR #343, opened 2016-07-15
Update travis scripts to install with pip

sbidoul:
This is a proof of concept to illustrate how installing with pip can help with dependency management.

There is no need for oca_dependencies.txt anymore as all dependencies are expressed in the `__openerp__.py`, possibly refined by `setup.py`, itself refined by `requirements.txt` if present.

This is IMHO a important enabler to progressively move towards smaller repos.

An example build with this PR is here: https://travis-ci.org/acsone/account-financial-reporting/builds/144847025 (including cross-repo dependencies).

If there is interest, I can finish this PR (ie update mqt test suite mainly).

### DIFF
--- a/travis/travis_install_nightly
+++ b/travis/travis_install_nightly
@@ -67,20 +67,41 @@ fi
 
 # Workaround to force using system site packages (see https://github.com/Shippable/support/issues/241#issuecomment-57947925)
 rm -f $VIRTUAL_ENV/lib/python2.7/no-global-site-packages.txt
-pip install --upgrade pip
+
+echo "Installing MQT requirements"
+pip install --upgrade "pip>=8.1.0"
 pip install --user -q --no-binary pycparser -r $(dirname ${BASH_SOURCE[0]})/requirements.txt
 pip install -q QUnitSuite coveralls codecov
 
 # Use reference .coveragerc
 cp ${HOME}/maintainer-quality-tools/cfg/.coveragerc .
 
-echo "Getting addons dependencies"
-clone_oca_dependencies
-clone_result=$?
-if [ "$clone_result" != "0"  ]; then
-    echo "Error cloning dependencies"
-    exit $clone_result
-fi;
+if [ -f "oca_dependencies.txt" ]; then
+    echo "Getting addons dependencies"
+    clone_oca_dependencies
+    clone_result=$?
+    if [ "$clone_result" != "0" ]; then
+        echo "Error cloning dependencies"
+        exit $clone_result
+    fi
+else
+    echo "Installing Odoo"
+    pip install -e ${HOME}/${REPO[1]}-${VERSION}
+    echo "Installing addons and dependencies"
+    pip install setuptools-odoo
+    setuptools-odoo-make-default -d .
+    for addon in $(ls setup/ -I README) ; do
+        echo "-e ./setup/$addon" >> requirements.txt
+    done
+    cat requirements.txt
+    # use --pre to get .devN versions from the wheelhouse
+    pip install --pre -r requirements.txt -f https://wheelhouse.odoo-community.org/oca
+    install_result=$?
+    if [ "$install_result" != "0"  ]; then
+        echo "Error installing addons and dependencies"
+        exit $install_result
+    fi
+fi
 
 echo "Install webkit (wkhtmltopdf) patched"
 (cd ${HOME}/maintainer-quality-tools/travis/ && wget -qO- http://download.gna.org/wkhtmltopdf/0.12/0.12.3/wkhtmltox-0.12.3_linux-generic-amd64.tar.xz | tar -xJ --strip-components=2 wkhtmltox/bin/wkhtmltopdf)
@@ -99,5 +120,13 @@ ls -l ${HOME}
 
 echo "Content of ${HOME}/dependencies:"
 mkdir -p ${HOME}/dependencies && ls -l ${HOME}/dependencies
+echo "pip installed packages:"
+echo VIRTUAL_ENV=$VIRTUAL_ENV
+echo PATH=$PATH
+which pip
+which python
+pip --version
+python -c "import sys; print sys.path"
+pip list
 
 set +e
--- a/travis/travis_install_nightly
+++ b/travis/travis_install_nightly
@@ -76,7 +76,7 @@ pip install -q QUnitSuite coveralls codecov
 # Use reference .coveragerc
 cp ${HOME}/maintainer-quality-tools/cfg/.coveragerc .
 
-if [ -f "oca_dependencies.txt" ]; then
+if [ -f "oca_dependencies.txt" ] || [ $(echo "$VERSION <= 7.0" | bc) -eq "1" ]; then
     echo "Getting addons dependencies"
     clone_oca_dependencies
     clone_result=$?
--- a/travis/travis_install_nightly
+++ b/travis/travis_install_nightly
@@ -94,8 +94,8 @@ else
         echo "-e ./setup/$addon" >> requirements.txt
     done
     cat requirements.txt
-    # use --pre to get .devN versions from the wheelhouse
-    pip install --pre -r requirements.txt -f https://wheelhouse.odoo-community.org/oca
+    # use --pre to get .devN versions
+    pip install --pre -r requirements.txt
     install_result=$?
     if [ "$install_result" != "0"  ]; then
         echo "Error installing addons and dependencies"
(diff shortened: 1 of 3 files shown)


moylop260:
- Is possible consider auto-create the `setup` folder when this one is not exists?

  Many people are using MQT to build theirs projects but we will require to create the `setup` folder and this matter maybe is not welcome.


- After merging what file we will need delete from our repositories? (oca_dependencies.txt, requirements.txt...)


- 
  > There is no need for `oca_dependencies.txt` anymore as all dependencies are expressed in the __openerp__.py

 What about if I want use a module from my custom repo instead of official repo? 
 What about use a dummy PR where is want to use a developer branch of a dependency to test it with other project e.g. reporting-engine-dev-x testing from account-financial-reporting?

sbidoul:
@moylop260 thanks for looking into this.

The setup directory is easily created with `setuptools-odoo-makedefault -d .` (part of setuptools-odoo package). For OCA, it's built nightly by the git bot if it does not exist yet.

If you want to work with a specifc branch of a dependency instead of getting it from wheelhouse.odoo-community.org, you just add it to the `requirements.txt` file, for example
`git+https://github.com/acsone/reporting-engine.git@somebranch#subdirectory=setup/report_xslx#egg=odoo-addon-report_xslx` 

You can even use different branches for different addons in the same repository. 

You can also declare other places to look for dependencies by adding `--find-links https://mywheelhouse/...` in the same requirements.txt.

So it's very flexible.

moylop260:
In principle, look good to me
Could you list the change that we will need execute in all oca projects after merging?

sbidoul:
@moylop260 there is nothing to change in OCA projects. We can remove oca_dependencies.txt when we are satisfied everything works fine.

One limitation is that it currently does not work with Odoo 7.0 because
- [ ] Odoo 7.0 does not pip install at the moment (https://github.com/odoo/odoo/pull/12816)
- [ ] setuptools-odoo does not support 7.0 at the moment (https://github.com/acsone/setuptools-odoo/commit/e39c831f69168bf96239aab41e1b3e73ae398193)
- [ ] 7.0 OCA addons versions are not normalized

Alternatively, I could preserve the current behaviour with oca_dependencies for 7.0, or use the new behaviour only if oca_dependencies.txt is not present.

moylop260:
I got it
Thank you

I like use pip if oca_dependencies is not present

sbidoul:
> I like use pip if oca_dependencies is not present

Ok, I will do that, it allows for a progressive deployment

dreispt:
It's a great idea to keep using the oca_dependencies.txt as now, and using pip installation when that file is not available.
I'll be using the dependencies logic for a the sattelite repo PoC (#338), so I need that code to be kept even if we use pip installation for dependencies.

sbidoul:
> use pip if oca_dependencies is not present

That mean pip will be used for all repos that have no oca_dependencies.txt today. There are many of them. I'm personally fine with that. I'll also need to disable the pip approach for 7.0 as we don't build wheels for 7.0 yet.

dreispt:
> That mean pip will be used for all repos that have no oca_dependencies.txt today. 

Yes, but I would expect for no Odoo module to be actually pip installed in those repos: if you have no oca_dependencies.txt, you're not using any modules outside the repo.

If you want to play safe, maybe you can use pip install only if a requirements.txt file is present, even if empty...

sbidoul:
This one should be good now. The new behavior is triggered if oca_dependencies.txt is absent.

dreispt:
AFAICS when we have an "oca_dependencies.txt" in the repo, behaviour is unchanged.
If it is missing, we then setuptools package the repo addons, and then pip install Odoo and these addons.

Questions:
a) I assume that this will seamlessly handle Python dependencies declared in the manifests, right?
b) What happens if our pip-installed addons have dependencies from other repos that are not packaged for pip installation?

sbidoul:
> a) I assume that this will seamlessly handle Python dependencies declared in the manifests, right?

Yes, setuptools-odoo does that.

> b) What happens if our pip-installed addons have dependencies from other repos that are not packaged for pip installation?

OCA addons can only depend on OCA addons. All OCA repos are packaged for pip installation automatically by a script running nightly. 

An improvement I'm considering is to have travis push to wheelhouse.odoo-community.org on success, so we don't have to wait the next day in case we have new cross-repo dependencies.

sbidoul:
One thing that remains TODO here is to disable the pip install behavour for Odoo versions <= 7.

lmignon:
@sbidoul May be this change is enough... https://github.com/acsone/maintainer-quality-tools/commit/01d3c99de5f8649fa51cb6cab443548de83bd3e6

pedrobaeza:
You should also do the same for lower versions (6.1, or even 5.0, 4.2 as we have in OCA/l10n-spain).

lmignon:
@pedrobaeza I'm not aware of projects using travis on branches < 7.0  In OCA/l10n-spain travis is not configured on 6.1, or even 5.0, 4.2 )

pedrobaeza:
OK, you're right. Transifex is not mandatory for that old ones.

sbidoul:
I pushed the check for versions <= 7.0. I hate bash.

dreispt:
I have a concern:
The MQT should be an universal tool, and not for exclusive use of the OCA.
For example, the design for the `oca_dependencies.txt` allowed shortcuts for the OCA, but took care that any arbitrary URL could also be used, so that I can use it with my private repos.

We should have this non-OCA use case in mind in the design.
Maybe adding a directive to the `.travis.yml` to explicitly enable pip installation?

lmignon:
@dreispt `requirements.txt`allow you to do the same thing... 
For exemple, in my requirements.txt used by alfodoo I specify the branch to use for the addon `cmis`
https://github.com/acsone/alfodoo/blob/9.0-add-travis/requirements.txt
(In alfodoo, `.travis.yml`use this branch of MQT https://github.com/acsone/alfodoo/blob/9.0-add-travis/.travis.yml)

sbidoul:
@dreispt oca_dependencies.txt sounds very oca-ish :wink: 

Joke aside, I think all the flexibility is there:
- if oca_dependencies.txt is present, pip is not used for installation, so you fallback on the current behaviour
- if you need to point to another wheelhouse in addition to OCA's you can add --find-links lines in requirements.txt
- as Laurent says you can point to individual addons branches that have a setup.py in requirements.txt too

Unless you want to use both pip and oca_dependencies.txt at the same time?

dreispt:
@sbidoul I got it now: the script is *adding* lines to the `requirements.txt`, not *replacing* it. So you can manually specify sources for packages as you just said.

> as Laurent says you can point to individual addons branches that have a setup.py in requirements.txt too
> Unless you want to use both pip and oca_dependencies.txt at the same time?

I think we should have this. For OCA is not an issue, but non-OCA users, they shouldn't be forced to make their repos pip-installable. In some wild cases, they might not even control the dependency repos to enforce that.

blaggacao:
So resuming the tooling, we will see in this space in the future:
- `pip` (this PR and odoo-setuptools).
  - theoretical advantage: formalized release cycle through packaging, pip ubiquity
  - theoretical disadvantage: special tooling for packaging required, private code requires _additional_ dedicated infrastructure
- `git` + `mv` (oca dependencies).
  - theoretical advantage: home made and well known
  - theoretical disadvantage: entry barrier (although it's actually dead simple, but the fact of _sounding_ unknown and specialized scares away). Somewhat a replica of `git submodule` in bash.
- `git submodules`
  - theoretical advantage: `git`, 
  - theoretical disadvantage: somewhat complicated
- `git subtree`
  - theoretical advantage: `git`, many say easier than submodule
  - theoretical disadvantage: huge repositories due to multiple histories in single repo, well odoo shallow is already 110MB, so it's relative.

We use `git submodule` as it is as agnostic as it can be. For us: `agnostic == strategic; True`
Never the less @sbidoul I like the suggestion!

sbidoul:
I updated this PR for odoo 10.0 and pypi support.

And example build of account-financial-reporting without `oca_dependencies.txt` is here: https://travis-ci.org/OCA/account-financial-reporting/jobs/229481888#L403

I think it is ready to merge now.

cc/ @dreispt

travis/travis_install_nightly
```diff
@@ -91,13 +112,17 @@ echo "Install webkit (wkhtmltopdf) patched"
 #     |___ <OdooRepo>-<Branch>/         <-- Odoo Server
 #     |___ maintainer-quality-tools/
 #     |___ build/<Owner>/<TestedRepo>/
-#     |___ dependencies/<DependencyRepo1>/
-#     |___ dependencies/<DependencyRepo2>/
```
dreispt: These dependency dirs will still be there if we have an `oca_dependencies.txt`, so IMO we should keep this.

sbidoul:
@dreispt indeed, that was a remnant of the initial prototype. Fixed and rebased.

bealdav:
nice feature, thanks a lot for this work :+1:

sbidoul:
superseded by #646

PR #351, opened 2016-09-10
[IMP] coveragerc: Ignore NotImplementedError

lasley:
This PR adds `NotImplementedError` to ignores in `.coveragerc` - which is generating what I consider to be false positives in some modules. Seems a bit redundant to be testing something that is clearly not implemented IMO

### DIFF
--- a/cfg/.coveragerc
+++ b/cfg/.coveragerc
@@ -18,3 +18,5 @@ exclude_lines =
     pragma: no cover
     # Don't complain about null context checking
     if context is None:
+    # Don't check obviously not implemented
+    raise NotImplementedError


pedrobaeza:
But this will work when you put an error message like `raise NotImplementedError('Error')`?

For now, I put the `# pragma: no cover` on those sentences.

moylop260:
👍  Useful for abstract classes and tests implemented on sub-class

lasley:
@pedrobaeza - yup this should work when you supply an error message. The exludes lines are regexes, so as long as the beginning of the line matches, the whole will be ignored. That's my understanding at least, and seemed to play out in the connectors.

pedrobaeza:
OK then. Let's add it!

pedrobaeza:
:+1:

yvaucher:
:+1:

Issue #348, opened 2016-09-06
Travis is green on catastrophic failures


lasley:
When there is a catastrophic issue with core, such as an ImportError, Travis is displaying a false green. I believe this is because the test runner isn't actually initialized yet, so we don't have any output.

Example - https://travis-ci.org/laslabs/odoo-connector-carepoint/jobs/157901951#L983

I'm assuming Odoo is throwing an exit code, I wonder if it would be possible to catch this and fail if anything other than a `0`?

osvalr:
@lasley I confirm this is happening.

moylop260:
Maybe is time to use the regex of runbot to detect errors on logs https://github.com/odoo/odoo-extra/blob/8.0/runbot/runbot.py#L42

lasley:
@moylop260 - I feel like the existing regex is accomplishing the same thing though, just with more options. It's definitely syntactically the same.

I took a look at the code, and I'm pretty sure it's because we are only listening for `CRITICAL` log level on versions >7. All of my samples are failing with an `ERROR` log level.

I proposed a fix in #352 but I have no idea how to test it. Can you point me in a direction?

pedrobaeza:
Closing this as there's an ongoing PR

PR #352, opened 2016-09-15
[FIX] test_server: Add ERROR log level

lasley:
* Also listen for ERROR log level on versions >7

#348

### DIFF
--- a/travis/test_server.py
+++ b/travis/test_server.py
@@ -30,7 +30,7 @@ def has_test_errors(fname, dbname, odoo_version, check_loaded=True):
         'failed sending mail',
         ]
     errors_report = [
-        lambda x: x['loglevel'] == 'CRITICAL',
+        lambda x: x['loglevel'] in ['CRITICAL', 'ERROR'],
         'At least one test failed',
         'no access rules, consider adding one',
         'invalid module names, ignored',


moylop260:
You need increase the number of EXPECTED ERROR for this first build from .travis.yml file in order to consider the new type of error.

pedrobaeza:
The problem with this is that sometimes we test that there are errors and that the errors raise, like this case: https://travis-ci.org/OCA/server-tools/jobs/166895188#L1045. Maybe tweaking EXPECTED_ERRORS can work for these cases.

lasley:
Why are we not catching the errors with `self.assertRaises()`? Or are these simply `logging.error` calls?

pedrobaeza:
They are caught that way, but this doesn't prevent the ERROR message in the log.

lasley:
That seems to only be in instances where a logger is specifically outputting an error message via `logger.error`. The following example generates no log:

```
>>> import unittest
>>> 
>>> def error():
...     raise Exception('Derp')
... 
>>> class MyTest(unittest.TestCase):
...     def test(self):
...         with self.assertRaises(Exception):
...             error()
... 
>>> unittest.main()
.
----------------------------------------------------------------------
Ran 1 test in 0.000s

OK
```

Regardless, that means we're on fairly thin ice with this solution in terms of scalability IMO. 

Is there ever an instance when log output is actually relevant to a test? It seems weird to me that we would test a log output, so what if we were to just raise the Odoo log level to `CRITICAL` instead of `ERROR`? This shouldn't affect tracebacks from real errors, it will only suppress superfluous logging calls meant to be read by a user.

pedrobaeza:
The problem is that you don't have control over that message that is spit out by Odoo. You only control the exception with `assertRaises`. Anyway, we can add the expected errors variable to control this. @moylop260, what do you think?

moylop260:
The case of example using `with self.assertRaises(Exception)` still show log.error from odoo orm.

Example:

```
import unittest
import odoo

def error():
    raise Exception('Derp')

class MyTest(unittest.TestCase):
     def test(self):
       with self.assertRaises(Exception):
             odoo.method()  # This method have a log.error('msg') and the log show it
```

We can use mute_logger in order to avoid show it.
```
import unittest
import odoo

def error():
    raise Exception('Derp')

class MyTest(unittest.TestCase):
     @odoo.mute_logger('logger.my_logger')  # Mute logger to avoid show the error in the log
     def test(self):
       with self.assertRaises(Exception):
             odoo.method()
```

[Example of this case](https://github.com/odoo/odoo/blob/f4d7d04c1d8dc8c46f81c1b2c0024fd5ec9ad190/addons/website_forum/tests/test_forum.py#L11-L16)

For the [sample modules of MQT](https://github.com/OCA/maintainer-quality-tools/tree/master/tests/test_repo) we need the test case to validate that the errors are processed correctly, then we shouldn't use a mute.

We are using a counter of errors.
Then if you increase the counter (for `CRITICAL` in this case) we need increase the expected value in order to avoid a MQT with that case tested and a CI green.

Issue #356, opened 2016-10-03
ImportError: No module named click


simahawk:
If you try to run `travis/run_pylint.py`

simahawk:
@nhomar move your a**!

moylop260:
Could you try install:  `pipi install pylint-odoo` in orde to fix it?
(its is locally, rigth?)

pedrobaeza:
If pylint-odoo is not in the requirements.txt of MQT, then the error is correct.

moylop260:
WIP from https://github.com/OCA/maintainer-quality-tools/pull/500

PR #357, opened 2016-10-03
[imp] install lxml via pip

simahawk:
`pip install lxml` is working today ;)

/cc @nhomar

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -2,3 +2,4 @@ simplejson
 pyserial
 pyyaml
 coveralls
+lxml
--- a/requirements.txt
+++ b/requirements.txt
@@ -3,3 +3,4 @@ pyserial
 pyyaml
 coveralls
 lxml
+click
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,7 +15,6 @@ addons:
   apt:
     packages:
       - expect-dev  # provides unbuffer utility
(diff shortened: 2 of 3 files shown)


requirements.txt
```diff
@@ -2,3 +2,5 @@ simplejson
 pyserial
 pyyaml
 coveralls
+lxml
+click
```
moylop260: Click?

simahawk: click is related to #356 
Unfortunately I didn't used to separated branches BUT @nhomar knows how to split these commits on merge, and he told me it was fine for such a small change :)

moylop260: We have installed click from install_nightly script when is required.
Then we dont need it here.

IMHO here should be just odoo requirements

simahawk:
​the issue is that if you use this on your local machine​ it won't work :)

moylop260:
Using the same docker image used by travis to compare the time:
 - `docker run -it --rm quay.io/travisci/travis-python:latest /bin/bash -c "time apt-get install python-lxml"`
  - 9 seconds

 - `docker run -it --rm quay.io/travisci/travis-python:latest /bin/bash -c "time pip install lxml"`
  - 52 seconds

screen shot 2016-10-04 at 2 12 23 pm

pedrobaeza:
So it seems to not be the same speed. I close the PR then.

PR #358, opened 2016-10-04
[ADD] travis: Add server_options

lasley:
This PR adds additional server_options flag in order to allow the ability to pass additional options to the server init, outside of the standard Travis options.

This is required in order to allow Runbot-Travis2Docker to pass the necessary [`--log-db`](https://github.com/odoo/odoo-extra/blob/master/runbot/runbot.py#L897) command to the underlying server. Using the pre-existing options would allow for overwrite via the Travis file, which is undesired.

* Add `SERVER_OPTIONS` env var
* Append server_options to relevant build commands

### DIFF
--- a/travis/test_server.py
+++ b/travis/test_server.py
@@ -181,7 +181,7 @@ def get_test_dependencies(addons_path, addons_list):
 
 def setup_server(db, odoo_unittest, tested_addons, server_path,
                  addons_path, install_options, preinstall_modules=None,
-                 unbuffer=True):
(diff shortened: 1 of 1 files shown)


moylop260:
👍

alan196:
:+1:

moylop260:
The red is unrelated

gurneyalex:
:+1: (a bit late, but provided live during odoo experience / code sprint)

Issue #361, opened 2016-10-05
Strange error on Transifex stage of Travis


pedrobaeza:
There's an strange error in this build:

https://travis-ci.org/OCA/product-attribute/jobs/165369928

@moylop260, do you have any clue?

moylop260:
Fixed https://github.com/OCA/maintainer-quality-tools/pull/363

PR #363, opened 2016-10-06
[FIX] test_server: Fix list concatenation with Null

moylop260:
Fix https://github.com/OCA/maintainer-quality-tools/issues/361


pedrobaeza:
I fast-track because this is preventing Transifex stage to work, and the error is not really happening.

PR #368, opened 2016-10-11
[REF] .travis.yml: Update global VERSION 9.0 to 10.0

moylop260:
In order to test latest version of odoo

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,7 +23,7 @@ virtualenv:
 env:
 
   global:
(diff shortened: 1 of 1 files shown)


pedrobaeza:
Shouldn't we test each of the Odoo versions, the same as @hbrunn has made here with openupgradelib: https://github.com/OCA/openupgradelib/pull/59/files ?

moylop260:
Good point!

PR #372, opened 2016-10-13
[IMP] coveragerc: Exclude ImportError guard from coverage

lasley:
* Exclude guard for import error from coverage reports


moylop260:
Good catch!
👍

PR #377, opened 2016-10-20
Add pylint-odoo dependency, used for pylint checks

### LINKED ISSUE #356: ImportError: No module named click

dreispt:
Fixes #356


moylop260:
For me is a good idea have lxml and pylint-odoo in requirements.txt for get all dependencies in local cases but we know that is slow use it for travis.

Could you add a sed to delete from requirements.txt the pylint-odoo
https://github.com/OCA/maintainer-quality-tools/blob/master/travis/travis_install_nightly#L10?

If you compare the time waste for this pr vs stable one you will a increase.

moylop260:
Other reason to add the sed is that we have separated the environments of lint and test
Then we dont need have pylint-odoo for test build and we dont need install all dependencies of test for lint one

dreispt:
@moylop260 sorry, I didn't understand - you dislike the PR?

moylop260:
I like the pr but plus a change.
We need faster builds.

Then you need add a sed command to delete the line added.

I mean, the requirements.txt for install test dont should have `pylint-odoo` then remove it with a sed command.

IMHO a full requirements is for local matter. Rigth?

dreispt:
@moylop260 Is the delay added significant?
If so, how about adding a `requirements-lint.txt` file instead. Builds could install it when in case the LINT flag is activated.

moylop260:
I like `requirements-lint.txt` for add flake8 and pylint-odoo

dreispt:
@moylop260 Travis builds are able to run without installing `pylint-odoo`?

lasley:
@dreispt The tests are, yeah. Pylint is only required for LINT=1 IIRC

dreispt:
Close?

nhomar:
I just update the PR to be mergeabele, is this valid yet @moylop260 ¿?

moylop260:
pylint needs hacks to install for travis, because we dont want lxml installable from pip requirements and we dont need it for no lints builds

PR #381, opened 2016-11-09
[FIX] Graceful coverage failures

lasley:
* Guard coverage calls to fail gracefully - both libraries output the error via logging, we don’t want to exit

Without this if coveralls fails due to no project existing, CodeCov will not be attempted. This means that BOTH Coveralls and CodeCov accounts are required to output coverage only via CodeCov.

I went ahead and guarded both though, because I didn't see any reason why not.


pedrobaeza:
Merging as 2 MQTs maintainers have approved it.

lasley:
Thanks for the quick merge, confirmed this works now too 👍  https://github.com/clouder-community/clouder/pull/154#issuecomment-259343780

PR #382, opened 2016-11-09
Recursive module search

jcdrubay:
### Objective
Encourage the use of maintainers quality tool when working on any project, not only OCA.

At Trobz (and I know for sure that we are not the only one), we use this method:
- One repo that contains all the code necessary for the project
- Part of this repo are subtrees (others can use sub-modules) for the common repositories of OCA or common internally
- The final repository structure is:
for a complete project with the below structure:
  - Project repo
    - addons
      - hr: oca hr repository as subtree / sub-modules
      - server-tools: oca server-tools repository as subtree / sub-modules
      - my_project_modules
        - my_module_1
        - my_module_2
    - odoo: subtree / sub-modules of odoo source code

### Issue faced
Maintainer quality tools assume that all the python modules are located directly in the repository.

Something like this:

- hr: the oca repository
  - hr_xxx: the modules

Tools does not discover modules located in sub-sub-folders or deeper, instead you get this message `Python modules not found`.

Note: This problem is specific to pylint tests, flake 8 tests run beautifully discovering all python code.

### Added Feature
Enable checks even when modules are not located directly in the root folder.

### Personal note
I am quite new to maintainers quality tool (my first commit here after using only few days) so I might miss something on how to use maintainer-quality-tool that would make this PR useless.

### DIFF
--- a/travis/run_pylint.py
+++ b/travis/run_pylint.py
@@ -3,13 +3,12 @@
 
 from __future__ import print_function
 
-import os
-import sys
-
 import click
 import pylint.lint
+import sys
 
 import getaddons
+
 
 CLICK_DIR = click.Path(exists=True, dir_okay=True, resolve_path=True)
 
@@ -24,31 +23,6 @@ def get_count_fails(linter_stats, msgs_no_count=None):
         linter_stats['by_msg'][msg]
         for msg in linter_stats['by_msg']
         if msg not in msgs_no_count])
-
-
-def get_subpaths(paths):
-    """Get list of subdirectories
-    if `__init__.py` file not exists in root path then
-    get subdirectories.
-    Why? More info here:
-        https://www.mail-archive.com/code-quality@python.org/msg00294.html
-    :param paths: List of paths
-    :return: Return list of paths with subdirectories.
-    """
-    subpaths = []
-    for path in paths:
-        if not os.path.isfile(os.path.join(path, '__init__.py')):
-            subpaths.extend(
-                [os.path.join(path, item)
-                 for item in os.listdir(path)
-                 if os.path.isfile(os.path.join(path, item, '__init__.py')) and
-                 (not getaddons.is_module(os.path.join(path, item)) or
-                  getaddons.is_installable_module(os.path.join(path, item)))])
-        else:
-            if not getaddons.is_module(path) or \
-                    getaddons.is_installable_module(path):
-                subpaths.append(path)
-    return subpaths
 
 
 def run_pylint(paths, cfg, beta_msgs=None, sys_paths=None, extra_params=None):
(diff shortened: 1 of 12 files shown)


jcdrubay:
@lasley Thanks for your time taking a look to the PR. I tried to take into account your comment by using an environment variable `PYLINT_RECURSIVE` which can be re-used if we want in the travis file matrix. Now we should have this behavior:  "By default do not check recursively except for pre-commit".

Most of travis tests are failing so I wish to debug to understand. But those tests are not trivial to setup locally (at least they are not to me).

I was wondering if there was any other guideline than the code. Thanks in advance for your help.

travis/run_pylint.py
```diff
@@ -67,7 +41,9 @@ def run_pylint(paths, cfg, beta_msgs=None, sys_paths=None, extra_params=None):
     sys.path.extend(sys_paths)
     cmd = ['--rcfile=' + cfg]
     cmd.extend(extra_params)
-    subpaths = get_subpaths(paths)
+
+    subpaths = getaddons.get_subpaths(paths)
```
lasley: This isn't conditional on the environment variable. Travis is likely failing due to this https://travis-ci.org/OCA/maintainer-quality-tools/jobs/174463515#L327

jcdrubay: I am building knowledge on MQT!!!
I prepared a script to run the travis tests locally, hopefully this will help me to debug.
I will publish the scripts once I successfully pass travis tests in the PR as well.

jcdrubay:
I started to prepare this small script to run automatic tests of maintainers-quality-tools travis-like: https://github.com/trobz/mqt-test-runner/blob/master/run.py

At this time, it only supports the 4 PYLINT tests from travis.yml file.

jcdrubay:
In the master branch, I see that in the job 7, in the second pylint execution for the precommit script, we find `55 errors` (See [here](https://travis-ci.org/OCA/maintainer-quality-tools/jobs/174392860#L487) starting from line 487) .

Also, the number of errors is different from the number expected:
 - I can see 55 errors while number expected is 26

Note: I have the same results with the script I prepared to test locally [here](https://github.com/trobz/mqt-test-runner/blob/master/run.py) and those results make sense to me, but I don't understand why the test in master branch is passing.

lasley:
TBH this gets further into MQT than I understand, outside of the fact that the count exists and was somehow determined at one point. Maybe @moylop260 has advice; he's the master here

dreispt:
@jcdrubay I think I can see what you want to achieve.
IMO that can can be done with script running on top of the current logic, minimizing changes to the existing code.

Something like this: given a root path, recursively browse the directory tree looking for addon directories. Detecting as addons directory is already implemented by `getaddons.is_addons(path)`.
Having a list of addons paths, you can iteratively run all the tests for each of them.
We still need to  modify the test running to support a list of paths, instead of a single target dir.

dreispt:
I did a smaller implementation for what you describe, at #392.
I hope it does all that you need. 
@jcdrubay can you try it?

jcdrubay:
Thanks @dreispt, I tried #392 and added comments.

dreispt:
Close, since it is solved by the implementation at #392 ?

jcdrubay:
I am ok to close as well.

Note that even after #392 and #394, pylint tests are still not executed recursively from the root directory, but I believe it is a good thing as it would be too heavy anyway to test pylint for all the code of the project at each commit.

dreispt:
IMO it's probably a good idea to have pylint tests run recursively.
My advice it to tackle that in it's own PR, so that we can get a smaller (and friendlier ) diff.

blaggacao:
@dreispt You review style is refreshingly encouraging! Just an opportunity to give a 👍

PR #389, opened 2016-11-28
[REF] maintainer-quality-tools: Use eslint instead of jshint for javascript lints

moylop260:
More info: https://github.com/OCA/pylint-odoo/pull/97

- Fix exit early if you want test both: `LINT_CHECK="1" TESTS="1"`

TODO:
- [x] Update this branch after merge the CI fixes: https://github.com/OCA/maintainer-quality-tools/pull/390 to get a green
- [ ] Create other PR to update expected errors after merge this one and https://github.com/OCA/pylint-odoo/pull/97

### DIFF
--- a/tests/test_repo/broken_module/broken_example.js
+++ b/tests/test_repo/broken_module/broken_example.js
@@ -1,0 +1,7 @@
+$(document).ready(function () {
+    $('.example').each(function () {
+        var oe_website_sale = this;
+    }) /*missing semicolon*/
+    /*Use of console log*/
+    console.log("This is similar to a print");
+}) /*missing semicolon*/
--- a/tests/test_repo/broken_module/broken_example2.js
+++ b/tests/test_repo/broken_module/broken_example2.js
@@ -1,0 +1,7 @@
+/*Use of "+function" instead of "Number(function" */
++function ($) {
+  'use strict';
+  var var_1 = "value1";
+  var var_2 = "value2";
+};
+/* Newline required at end of file but not found */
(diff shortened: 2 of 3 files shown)


pedrobaeza:
Where is the .jslintrc file?

moylop260:
Oh yeah, you are right!
The file here is missing, by default will be used from `pylint-odoo`
If we need change a particular check for MQT we can think in change the default file or add a custom here.

pedrobaeza:
OK, I would prefer to bundle here one for being less obscure than in the python library, but it's nothing critical.

lasley:
> I would prefer to bundle here one for being less obscure than in the python library,

I like this too, but how would we maintain sync between the two if we did do this?

pedrobaeza:
pylint_odoo maybe shouldn't include any sample configuration file.

lasley:
I admittedly didn't know it did until this conversation came up, so I think that's probably a good call. Keep all the sample files centralized here, which is also part of our builds so is included there in case we also want to use them as defaults.

The flipside is that we're making `pylint_odoo` dependent on MQT in a way, right?

pedrobaeza:
The only downside I see is that pylint_odoo can be self-checked unless you hardcode a configuration in the test.

moylop260:
Team
FYI we have all approves to merge https://github.com/OCA/pylint-odoo/pull/97
But we need merge this PR together to work without errors.

But we don't have approves here.
Tell me what I should change to receive your approval or give me your approval to enjoy this feature, please.

moylop260:
The approved https://github.com/OCA/pylint-odoo/pull/97 require this PR then I'll merge both but we will work changes about in other ones

PR #390, opened 2016-11-28
[FIX] Fixes CI result

moylop260:
- [REF] broken_module: Fix False expression, now is required a string
- [REF] flake8 rc: 'Skip F999 dictionary key 'name' repeated with different values' because pylint support this check
- [REF] test_server: Fix flake8


moylop260:
Now travis result is green... (But is red green because there is a issue related with travis-ci: https://github.com/travis-ci/travis-ci/issues/6943)

moylop260:
cc @dreispt @pedrobaeza @lasley @alan196 @jjscarafia

PR #391, opened 2016-11-28
.travis.yml: sass install instructions

pedrobaeza:
Include instructions for installing sass command.

### DIFF
--- a/sample_files/.travis.yml
+++ b/sample_files/.travis.yml
@@ -1,5 +1,7 @@
 language: python
 sudo: false
+# you need to enable sudo for installing sass command (needed for website related repos)
+# sudo: true
 cache:
   apt: true
   directories:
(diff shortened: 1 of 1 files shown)


pedrobaeza:
cc @lasley

sample_files/.travis.yml
```diff
@@ -1,5 +1,7 @@
 language: python
 sudo: false
+# you need to enable sudo for installing sass command (needed for website related repos)
+# sudo: true
```
lasley: No need for this actually, RVM handles it

sample_files/.travis.yml
```diff
@@ -66,9 +68,17 @@ env:
...
+#before_install:
+#  - rvm install ruby --latest
+#  - gem install bootstrap-sass
+#  - gem install compass --pre
+
 install:
   - git clone --depth=1 https://github.com/OCA/maintainer-quality-tools.git ${HOME}/maintainer-quality-tools
   - export PATH=${HOME}/maintainer-quality-tools/travis:${PATH}
+  # Enable this if your repositoru needs sass command (needed for website related repos)
+  #- sudo apt-get install ruby-sass
```
lasley: This isn't needed either, dependency is installed in `bootstrap-sass` into Ruby env

lasley: Just to note - I did benchmark the two against each other, and obviously the gem install was much slower especially given the Ruby compilation for RVM upgrade.

The downside of the `apt` package, outside of `sudo` required, is that it is incredibly out of date. This will likely not emulate a production server, which would be running more up-to-date assets (at least mine are)

pedrobaeza:
@lasley, is now correct?

PR #392, opened 2016-12-01
Allow tested directory to contain addon dirs

dreispt:
Previously the tested directory ought to be an addons directory.
With this change, it can instead contain several addon directories.
A search will be done inside it,to find all contained addon directories,
and test will be run on all of them.

Intended to replace #382

### DIFF
--- a/travis/test_server.py
+++ b/travis/test_server.py
@@ -129,9 +129,9 @@ def get_addons_path(travis_dependencies_dir, travis_build_dir, server_path):
     :param server_path: Server path
     :return: Addons path
     """
-    addons_path_list = get_addons(travis_dependencies_dir)
-    addons_path_list.insert(0, travis_build_dir)
-    addons_path_list.append(server_path + "/addons")
+    addons_path_list = get_addons(travis_build_dir)
+    addons_path_list.extend(get_addons(travis_dependencies_dir))
+    addons_path_list.append(os.path.join(server_path, "addons"))
     addons_path = ','.join(addons_path_list)
     return addons_path
 


jcdrubay:
Thanks @dreispt for your help.
Unfortunately, when committing, the pylint tests are skipped with the message `Python modules not found`.

Yet flake 8 tests are running well.

dreispt:
Yes, this PR only works on the server tests side.

We can work on the pylint side in a different PR.
I'm not sure exactly how you are running the tests.
test_pylint work on the current working dir, so you either change it before the checks, or modify test_pylint so that it can run on the build directory.
Have a look at the pointers in #394

In the meanwhile, can you approve this PR?

jcdrubay:
I approved but it seems a bit early no?
 - travis test fail: I guess that we have to fix it
 - missing pylint tests: ok to work on another PR, but we can update the description of this one?

I am running the test by committing files from my "complete" project having the structure described in #382 (.git repository is at the root). Or by executing manually as I mentioned in #394 : `TRAVIS_PULL_REQUEST="true" TRAVIS_BRANCH="HEAD" TRAVIS_BUILD_DIR=pwd -P.git/hooks/test_pylint`

> test_pylint works on the current working dir:
Ok, but the idea if #382 was to make it work recursively on sub-directories: `{my-project} > addons > oca-hr`.

However, in my context, I believe that the easiest and best way will be to not create a symlink for the file `pre-commit` but to copy it only changing one line:
`TRAVIS_PULL_REQUEST="true" TRAVIS_BRANCH="HEAD" TRAVIS_BUILD_DIR="$(dirname $0)/../../addons/{my_project_addons}" $(dirname $0)/test_pylint`  because normally, only the content of {my_project_addons} should be updated during developments.

`{my_custom_project_addons_path} = {my-project} > addons > my_project_modules`

Note: I believe that #382 was only covering the case when committing from the root directory.

dreispt:
Let's split the big problem in two smaller ones (and two smaller and easier to review PRs).

I believe this one solves the server tests.
Let's discuss the pylint needed changes at #394 or a new issue/pr.

I don't see that as a reason to block this, but otoh there's no need to rush it.

--dr

dreispt:
Oh, and Travis is green except for the INSTANCE_ALIVE build, which is unrelated.

dreispt:
Merge?

blaggacao:
@dreispt Thank you! It wasn't obvious to me, until I found this commit. I was about to open a new discussion...
Does this mean I am supposed / able to do the following:

`RUN_COMMAND_MQT_1: "rsync -av --delete --exclude 'i18n' /home/src/odoo-cc/* /home/testbed/odoo-9.0"`
`RUN_COMMAND_MQT_2: "rsync -av --delete --exclude 'i18n' /home/src/odoo-ee /home/testbed/dependencies/"`
`RUN_COMMAND_MQT_3: "rsync -av --delete --exclude 'i18n' /home/src/access_addons /home/testbed/dependencies/"`
`RUN_COMMAND_MQT_4: "rsync -av --delete --exclude 'i18n' /home/src/devco /home/testbed/build/"`
`RUN_COMMAND_MQT_4: "rsync -av --delete --exclude 'i18n' /home/src/devco-ee /home/testbed/build/"`

while `','.join(addons_path_list)` in this case guarantees the right order (for example for the odoo-ee web module override). But what guarantees can I count on for the get_addons ordering in the case that devco-ee is doing a module override on devco?

blaggacao:
@dreispt I guess I responded my question through https://github.com/OCA/maintainer-quality-tools/pull/408/files @moylop260 can you include this?

PR #396, opened 2016-12-06
test_pylint fallback its path on current working directory if none de…

jcdrubay:
…fined as option or environment variable TRAVIS_BUILD_DIR

### DIFF
--- a/travis/run_pylint.py
+++ b/travis/run_pylint.py
@@ -11,7 +11,6 @@ import pylint.lint
 
 import getaddons
 
-
 CLICK_DIR = click.Path(exists=True, dir_okay=True, resolve_path=True)
 
 
--- a/travis/run_pylint.py
+++ b/travis/run_pylint.py
@@ -3,13 +3,14 @@
 
 from __future__ import print_function
 
+import click
 import os
+import pylint.lint
 import sys
 
-import click
-import pylint.lint
-
 import getaddons
+import travis_helpers
+
 
 CLICK_DIR = click.Path(exists=True, dir_okay=True, resolve_path=True)
 
@@ -78,7 +79,7 @@ def run_pylint(paths, cfg, beta_msgs=None, sys_paths=None, extra_params=None):
 
 @click.command()
 @click.option('paths', '--path', envvar='TRAVIS_BUILD_DIR',
-              multiple=True, type=CLICK_DIR, required=True,
+              multiple=True, type=CLICK_DIR,
               help="Addons paths to check pylint")
 @click.option('--config-file', '-c',
               type=click.File('r', lazy=True), required=True,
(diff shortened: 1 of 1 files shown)


travis/run_pylint.py
```diff
@@ -3,13 +3,14 @@
 
 from __future__ import print_function
 
-import os
-import sys
-
 import click
+import os
 import pylint.lint
+import sys
```
dreispt: Imports from the standard library should come before imports from third party packages.
See https://github.com/OCA/maintainer-tools/blob/master/CONTRIBUTING.md#imports

jcdrubay:
I took into account the small comment from @dreispt and created #397 rebasing this PR to have a clean commit history.

dreispt:
You don't need to create a new PR.
Just add your commit to this branch and push it, and this PR will be updated.
Github allows us to squash the commits when merging, so you don't need to worry about that.

In case you care and do want to tidy your commit history, you can use `git commit --amend` to modify the last commit, instead of adding a new one.

travis/run_pylint.py
```diff
@@ -78,7 +80,7 @@ def run_pylint(paths, cfg, beta_msgs=None, sys_paths=None, extra_params=None):
 
 @click.command()
 @click.option('paths', '--path', envvar='TRAVIS_BUILD_DIR',
-              multiple=True, type=CLICK_DIR, required=True,
+              multiple=True, type=CLICK_DIR,
```
moylop260: What about ` default=os.getcwd()`?

dreispt:
LGTM

travis/run_pylint.py
```diff
@@ -10,6 +10,8 @@
 import pylint.lint
 
 import getaddons
+import travis_helpers
```
moylop260: Please, fix `./travis/run_pylint.py:13:1: F401 'travis_helpers' imported but unused`
https://travis-ci.org/OCA/maintainer-quality-tools/jobs/183560210#L468

travis/run_pylint.py
```diff
@@ -97,6 +100,7 @@ def main(paths, config_file, msgs_no_count=None,
     to check fails of odoo modules.
     If expected errors is equal to count fails found then
     this program exit with zero otherwise exit with counted fails"""
+
```
moylop260: If this change is not required then remove it

travis/run_pylint.py
```diff
@@ -78,7 +80,8 @@ def run_pylint(paths, cfg, beta_msgs=None, sys_paths=None, extra_params=None):
 
 @click.command()
 @click.option('paths', '--path', envvar='TRAVIS_BUILD_DIR',
-              multiple=True, type=CLICK_DIR, required=True,
```
moylop260: Could you use `required=True`?

jcdrubay:
Thanks @moylop260 and @dreispt for your help. :)

PR #399, opened 2016-12-21
[REF] requirements: Add xlsxwriter package

moylop260:
New package requirement from [odoo#10.0](https://github.com/odoo/odoo/blob/81a4830c738b3c601a3aae7f7bcdcd7a4f7abb1f/requirements.txt#L40)

### DIFF
--- a/travis/requirements.txt
+++ b/travis/requirements.txt
@@ -36,3 +36,5 @@ vatnumber
 vobject
 werkzeug
 xlwt
+# odoo-10.0
+XlsxWriter


dreispt:
Why don't we  pip install the requirements.txt after cloning the Odoo project?

moylop260:
Maybe for historical reasons.
Older version of odoo didn't have requirements.txt
Your idea is a good approach.
Could you create the issue and add the tag "needs help"?

We will need:
 - Remove lxml with a sed command because is installed from apt. (Maybe other package with similar case)
 - Add tests case for all versions.
 - Use alternative requirements if there is not present in that version
 - Check if a package is installed from main image of travis to remove it
 - Check if a package require developer tools package to compile it

However, if currently we don't have the candidate to work on this behaivour this pr helps to avoid add in all requirements.txt of oca projects for version 10.0 this package manually.

lmignon:
@moylop260 https://github.com/OCA/maintainer-quality-tools/pull/343 uses the requirements.txt from Odoo...

moylop260:
Maybe not for all cases
Here you see a installation from mqt yet:

https://travis-ci.org/OCA/maintainer-quality-tools/jobs/167031005#L232

lmignon:
@moylop260 good catch.. I was pretty sure that was the case in this PR.  Internally we always pip install every modules in our projects and for odoo we use `pip install -r https://raw.githubusercontent.com/odoo/odoo/9.0/requirements.txt' to install the official python requirements.

moylop260:
Yes we do

osvalr:
:+1:

PR #400, opened 2016-12-23
[REF] git-hook: Ask if we want run lints

moylop260:
[REF] pre-commit: when TRAVIS env var is true, skip question and run lints

Show a friendly confirmation message in order to skip easiest the lint check in each commit:
screen shot 2016-12-22 at 10 10 03 pm

### DIFF
--- a/git/pre-commit
+++ b/git/pre-commit
@@ -3,11 +3,11 @@
 if [ "$NOLINT" ]; then exit 0; fi
 
 if [ "$TRAVIS" != true ]; then
-	read -p "Do you want to run lints? (y/N): " yn </dev/tty
-	case $yn in
-		[Yy]* ) true;;
-		* ) exit 0;;
-	esac
+    read -p "Do you want to run lints? (y/N): " yn </dev/tty
+    case $yn in
+        [Yy]* ) true;;
+        * ) exit 0;;
+    esac
 fi
 
 set -v
--- a/git/pre-commit
+++ b/git/pre-commit
@@ -2,6 +2,14 @@
 
 if [ "$NOLINT" ]; then exit 0; fi
 
+if [ "$TRAVIS" != true ]; then
+	read -p "Do you want to run lints? (y/N): " yn </dev/tty
+	case $yn in
+		[Yy]* ) true;;
+		* ) exit 0;;
+	esac
+fi
+
 set -v
 
 FLAKE8_CONFIG_DIR="$(dirname $0)/cfg"
--- a/git/pre-commit
+++ b/git/pre-commit
@@ -1,8 +1,8 @@
 #!/bin/bash
 
-if [ "$NOLINT" ]; then exit 0; fi
+if [ "$NOLINT" == "1" ] || [ "$NOLINT" == true ]; then exit 0; fi
 
-if [ "$TRAVIS" != true ]; then
+if [ "$TRAVIS" != true ] && [ -z ${NOLINT+x} ]; then
     read -p "Do you want to run lints? (y/N): " yn </dev/tty
     case $yn in
         [Yy]* ) true;;


pedrobaeza:
You don't need this, because there's already a mechanism: put `NOLINT=1 git commit ...`.

An interactive question is not allowed for some automatic process (like git integration in IDEs), so :-1: for this

moylop260:
> An interactive question is not allowed for some automatic process (like git integration in IDEs)

For Git integration in IDE you should use `git commit --no-verify ...`.

We have 3 options for this case:
1. git --no-verify
2. environment variable to avoid send `--no-verify` in each command.
3. Prompt question, if you forgot set a environment variable or set no-verify, you have the last option of cancel the large process in each commit.

moylop260:
Or maybe, if we have a explicit defined NOLINT="1" (instead of a undefined) then never ask.
What do you think?

osvalr:
I agree with this just to not use `NOLINT=0` as prefix everytime a `git commit` is typed, 
So :+1:

git/pre-commit
```diff
@@ -2,6 +2,14 @@
 
 if [ "$NOLINT" ]; then exit 0; fi
 
+if [ "$TRAVIS" != true ]; then
```
moylop260: @tomeyro 
Could you consider the following cases?

| NOLINT | Ask before of run lints? | 
|-----------|------|
| Undefined       |  yes |
| 1       |  exit early (it is) |
| 0        | No  |

tomeyro: So, ask only when NOLINT is undefined?

moylop260: Yeah

tomeyro: Should I push the changes here @moylop260?

tomeyro:
Please @moylop260 check the commit and tell me if that is what you needed. When NOLINT is 1 or true the tests are skipped, if undefined, it will show a prompt to ask the user if they want to run the tests (if also TRAVIS is not true), and if any other value (preferably 0), it will skip the question and run the tests.

moylop260:
Based on the last table of my comment:

| NOLINT | Ask before of run lints? | 
|-----------|------|
| Undefined       |  yes |
| 1       |  exit early (it is) |
| 0        | No  |

Running `NOLINT="0" maintainer-quality-tools/git/pre-commit` (The last case) is asking yet.

tomeyro:
Running the script on commit with multiple cases (also note that I put a stop and just prints "TESTING.." when the tests are supposed to run):

captura de pantalla de 2016-12-26 13-46-57

Running the script directly with NOLINT on 1, 0, and undefined (I removed the stop here):

captura de pantalla de 2016-12-26 13-52-41

moylop260:
You are right!
👍

tomeyro:
:+1:

moylop260:
@pedrobaeza 
We find a way to work with both point of view, searching a middle point.
Could you review again, please?

zaoral:
+1

I think this is a good idea for the programmer point of view, my point of view. Using NOLINT=1 every time is a really cumbersome way to work, it is simple use just the regular git commit.

In my regular work day I use to set the environment variable NOLINT=1 as a global environment variable just to avoid add the variable every time I made git commit. However, there are times I want to run the lints, and times I do NOT want to run the lints. Set/unset this environment variable is still a 	comfortless task because I have to ask also almost every time what is the current status of the environment variable, and... when I skip this last validation... always end up running the lints when I had no intention of.

Having this question when we do git commit is for me the best way to let programmers have a better control and a more comfortable way to manage when to/not to run the lints.

If the guys found the way to have this functionality and also have a way to be able to run it automatically in scripts without any problem I think should be considere to be merge.

Best Regards

pedrobaeza:
OK, it looks fine to me, but please check Travis status for this PR. Have you answered `no` to the linter when committing? :stuck_out_tongue:

moylop260:
We have a travis-ci core issue with a false red: https://github.com/travis-ci/travis-ci/issues/6861

moylop260:
FYI you can see MQT issue case here: https://github.com/travis-ci/travis-ci/issues/6943
(The script bash modified is not affected by python lints :stuck_out_tongue: :stuck_out_tongue:)

pedrobaeza:
OK, you can proceed with the merge then

dreispt:
Remember that authors should refrain from merging their own PRs.

pedrobaeza:
That rule has no sense with the corresponding approvals IMO.

PR #401, opened 2016-12-31
[IMP] Switch git clone for archive download in depends download

lasley:
I was messing around in LasLabs/docker-odoo#1 & realized that an easy way for us to get a pretty good speed boost on our builds is to use Github archives instead of cloning the repo.

Is there a specific reason we are cloning here? Are these archives not always current or something? I tested a bit and they seem to be current.

I'm not proposing this code in its current form, but something very close. Any thoughts?

### DIFF
--- a/travis/clone_oca_dependencies
+++ b/travis/clone_oca_dependencies
@@ -56,14 +56,14 @@ def parse_depfile(depfile, owner='OCA'):
 
 def git_checkout(deps_checkout_dir, reponame, url, branch):
     checkout_dir = osp.join(deps_checkout_dir, reponame)
-    if not osp.isdir(checkout_dir):
-        command = ['git', 'clone', '-q', url, '-b', branch, checkout_dir]
-    else:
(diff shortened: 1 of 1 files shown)


lasley:
Understood thanks for the breakdown @moylop260. Sounds like the instance introspection plans will be fine if we switch this up to a `--single-branch --depth=1` though? I thought there was a bit of a speed decrease there, but my benchmarking confirms what you say.

dreispt:
Sounds good, but on the other hand, the day we make it easy to use MQT to run tests in our dev machines, I would like the option to git clone, even if that's  not the default.

lasley:
Updated to use `--single-branch --depth=1`

pedrobaeza:
Why Travis is red?

lasley:
Weird - it segfaulted on coverage. I rebuilt the failing one

> /home/travis/.travis/job_stages: line 57:  5955 Segmentation fault      (core dumped) coverage run --append ./travis/self_tests

PR #407, opened 2017-01-09
[DOC] on different testing environments

blaggacao:
Resulting of https://github.com/OCA/maintainer-quality-tools/pull/392#issuecomment-271409121 this is an attempt to open up this greate set of tools for a broader user base.

PR #408, opened 2017-01-09
[IMP] make addons path order predictable after #392

blaggacao:
of https://github.com/OCA/maintainer-quality-tools/pull/407/files#diff-04c6e90faac2675aa89e2176d2eec7d8R103


blaggacao:
@moylop260 @pedrobaeza If we could incorporate this, I would be a happier person the day after it has been merged. Independently of if my doc PR will be accepted or not.. 😄

moylop260:
Are you testing this change using [odoo/enterprise](https://github.com/odoo/enterprise)?
I mean, this special repository require defined before of all dependencies to work well!

blaggacao:
@moylop260 enterprises precisely is my use case, thanks for bringing this into the discussion. I'm testing locally. However, I chose to include enterprise module into the dependency folder, which is already put in the right order by `server_test.py`. The use case here is that you might have two versions of an addon, let's suppose:
- vauxoo-cc
 - l10n_mx (community version)
- vauxoo-ee
 - l10n_mx (enterprise version)

So I would do:
```
RUN_COMMAND_MQT_1: "rsync -av --delete /home/src/vauxoo-ee/* /home/testbed/build/1_vauxoo-ee"
RUN_COMMAND_MQT_2: "rsync -av --delete /home/src/vauxoo-cc/* /home/testbed/build/2_vauxoo-cc"
```
and voila, thank's to the sorting in this commit, 1_vauxoo-ee get's placed before 2_vauxo-cc in the addons path, and l10n_mx (enterprise version) is used in the tests (making use of the module override).

Without that, order is not very predictable (http://stackoverflow.com/questions/4813061/nonalphanumeric-list-order-from-os-listdir-in-python).

blaggacao:
@pedrobaeza This behaviour was intentionally included by @dreispt and @moylop260 in #392 
A valid use case is the envisioned repo split (which you can manage the way you want on that side, we manage it with submodules). Just think of an additional client repo, which you want to test with MQT, but without altering the existing "expected folder structure" too much, just as #392 implicitly confirms, this seems to be the way to go. Do you agree?

blaggacao:
@jcdrubay this could be interesting for your setup as well (being able to control addons path order after #392 ).

moylop260:
> have a commit message that clearly describes the purpose of the change.

...and a PR title

blaggacao:
@dreispt thanks for the approval. @moylop260 can you squash commit? Using something like the updated title should be fine, I guess?

blaggacao:
@moylop260 Would you consider being the third approver & merge, so we can relocate our attention?

travis/getaddons.py
```diff
@@ -71,7 +71,7 @@ def get_addons(path):
         res = [path]
     else:
         res = [os.path.join(path, x)
-               for x in os.listdir(path)
+               for x in sorted(os.listdir(path))
```
moylop260: Could you add a print-screen with returned values using the following `oca_dependencies.txt`:
`enterprise git@github.com:odoo/enterprise.git`

blaggacao: Let me ask a question: Are you insinuating that enterprise dependencies would get installed into the `odoo-9.0` folder? That would just break (because of the web module (we can't predict order reliably)), but this is then a problem of `oca_dependency` and unrelated here. But be assured, here is your answer: https://github.com/OCA/maintainer-quality-tools/blob/master/travis/test_server.py#L132-L134 `travis_dependencies_dir` comes first.
Again, within that folder, we cannot predict any order, so this commit is completely `pareto efficient`. Can you explain at least why do you think enterprise needs to come before other `travis_dependencies_dir` - addons. Yes, I'm aware that it needs to come before `odoo-cc` and why.

blaggacao:
@moylop260 Ok, will change the commit message. Sorry, as I had no answer, I assumed a quick squash-renaming through the github interface was no problem.

blaggacao:
@moylop260 Can you please help me with this:
```
λ git fetch origin pull/408/head:patch-2
fatal: Couldn't find remote ref pull/408/head
```
I'm not able to pull the pull request from github in order to force push afterwards....

dreispt:
We can change the commit message upon merge.
@moylop260 are you OK with merging?

travis/getaddons.py
moylop260: Then, you won't have problem with the print-screen, right?

blaggacao: I would need some non trivial amount of help, as I'm totally unaware of how to setup a repo structure to work with `oca_dependencies.txt` at the moment... It's for me pretty much out of context right now, so I do not understand quite well where to start and what printscreen exactly you are expecting? Would it be ok for you, if I give you the following answer:

Before:
```
dependencies/enterprise/
dependencies/repo123/
dependencies/abc1231/
```
--addons-path [ODER UNKOWN]


After:
```
dependencies/enterprise/
dependencies/repo123/
dependencies/abc1231/
```
--addons-path dependencies/abc1231,dependencies/enterprise,dependencies/repo123
[ALPHANUMERICAL]

Do you really need a printscreen for this? I think I don't understand, actually, sorry. I hardly belief that this is what you want to see:

```
>>> sorted(['b', 'c', 'a'])
['a', 'b', 'c']
```

But what else is your concern? That's the only change introduced. I do not get it. Can you please explain?

blaggacao: Why do you want to make me spent 1 hour or more to produce a print screen on something so obvious? Is this really necessary for you evaluation of this changeset (please do not burden upon me unrelated research questions).

blaggacao:
@moylop260 I do not know how to deliver the requested change. As @dreispt suggested, there is a easy solution to it. May I ask you to reconsider your verdict? I promise not to do changes from github directly any more. I'm already set up so won't happen any more, was just for the first few smallistic changes...

blaggacao:
@moylop260 any news on this? I don't have anything order pending on my side, waiting for approval... Or am I wrong?

moylop260:
I need time for a paranoid test.
In order to avoid request you a new tests or a new print-screen

If something help us to test it (not just code review with a print-screen or a unittest) I'm agree.
NOTE: The tests concept told us that If we change a `comment` we need run all tests again (paranoid maybe, but this project is too critical)

blaggacao:
@moylop260 great to hear that, I added some test, however am not too familiar with the ideas of testing here at MQT, could you have a look at it?

blaggacao:
**T r a v i s i s**
🍏 🍏 🍏

moylop260:
Thanks for add unittest
I see the log and I don't find the print added in the output.

I run a t2d and I applied the following patch:
```diff
diff --git a/travis/self_tests b/travis/self_tests
index 2df72cb..1e87fb3 100755
--- a/travis/self_tests
+++ b/travis/self_tests
@@ -43,6 +43,7 @@ addon_paths_alfanumerical_order_should = [
 ]
 combined = zip(addon_paths_alfanumerical_order_is,
                addon_paths_alfanumerical_order_should)
+import pdb;pdb.set_trace()
 print combined.__repr__()
 for ist, should in combined:
     assert ist.rstrip('/').rstrip('\\').endswith(should)
```

After that ran `self_tests` and I see the following output:
screen shot 2017-01-14 at 3 47 06 pm

The variable `addon_paths_alfanumerical_order_is` is empty.
Is right it is?

moylop260:
I tested directly this feature and works like as expected:
screen shot 2017-01-14 at 4 13 43 pm

I mean,
The order of the paths modules are:
1st. TRAVIS_BUILD_DIR
2nd. All dependencies modules from oca_dependencies (here are ordered)
3rd. Odoo dependencies

With this feature we can use from oca_dependencies the following configuration to force a order:
```
# oca_dependencies.txt
project-a URL 02_project-a
project-b URL 01_project-b
```
Generating the following folders:
 - ~/dependencies/
   - 02_project-a/
   - 01_project-b/
And will be ordered `01_project-b,02_project-a`

blaggacao:
@moylop260 Nice, that oca_dependency does have the naming feature. Then it's even more powerfull.

I guess the problem with the tests lies here: https://github.com/OCA/maintainer-quality-tools/pull/408/files#diff-9a8ed2b6abec6a7d86cbe4578c91aa45R16
`repo_dir` can be two things:
1. something arround `/build/travis/maintainer-quality-tools/` or
2. `./tests/test_repo/`
That _seems_ inconsistent, and I don't know the reason, but it tricked me into assuming that
`../test_repo_with_subfolders/` would yield the desired repo coming from `./tests/test_repo/`, but with `TRAVIS_BUILD_DIR` set, we come from something arround `/build/travis/maintainer-quality-tools/`

Should be the `repo_dir` folder more consistent, that is always refer to the repo main folder?

moylop260:
You are right that line is the problem

Using:
`repo_dir_with_subfolders = os.path.join(repo_dir, "tests", "test_repo_with_subfolders")`

The tests works!

screen shot 2017-01-14 at 4 54 29 pm

blaggacao:
@moylop260 can you review my last commit only which is intended to fix this doubt about `repo_dir`, and from where we come there... Now it should take the right folder as `repo_dir_with_subfolders`...

blaggacao:
https://cloud.githubusercontent.com/assets/6644187/21958671/6e6843f4-da79-11e6-8d7e-19beec25645c.png
Which isn't what I excpet, neither... Good to have tests 😄 and a helping hand 💚

moylop260:
Now I see the print from output log
https://travis-ci.org/OCA/maintainer-quality-tools/jobs/192005807#L579

`[('/home/travis/build/OCA/maintainer-quality-tools/tests/test_repo_with_subfolders/1_testfolder', '1_testfolder'), ('/home/travis/build/OCA/maintainer-quality-tools/tests/test_repo_with_subfolders/2_testfolder', '2_testfolder')]`

Thanks you!

blaggacao:
@moylop260  There was another one, the `zip()` could produce an erroneously bad test being green...
This is why you didn't see the print in the logs in the first place...
`zip([], ['a', 'b']) == []; True`

blaggacao:
Fixed in last commit of https://github.com/OCA/maintainer-quality-tools/pull/419

Issue #398, opened 2016-12-15
ValueError: could not convert string to float: master


moylop260:
Steps to reproduce it:
1. Assign environment variable `VERSION="master"` in a .travis.yml
2. Wait travis build

Result:
`ValueError: could not convert string to float: master`

Expected:
Working fine like as other version of odoo/odoo

moylop260:
The problem is the sentence [float(odoo_version)](https://github.com/OCA/maintainer-quality-tools/blame/406b428c97b35f36398c6a10d10ca57621449929/travis/test_server.py#L140) from https://github.com/OCA/maintainer-quality-tools/pull/364

@StefanRijnhart
Should we use a `script_name = 'openerp-server' if os.path.isfile('openerp-server') else 'odoo-bin'`?

lasley:
I prefer the try/except personally. That pattern is more pythonic

StefanRijnhart:
Fine with me!

blaggacao:
@moylop260 `try/except` and revert #364 seems more elegant, as it reduces some method's signatures
@lasley EAFP: European Association of Fishing Pathologists, is that what you mean?? 😄

dreispt:
@blaggacao  ["Easier to ask for forgiveness than permission."](http://stackoverflow.com/questions/11360858/what-is-the-eafp-principle-in-python#11360880)

nhomar:
I am considering this solved, re-open it if not please.

thanks.

PR #409, opened 2017-01-09
[FIX] "master" version

blaggacao:
Reverts OCA/maintainer-quality-tools#364 as discussed in https://github.com/OCA/maintainer-quality-tools/issues/398 add try except statement instead

### DIFF
--- a/travis/test_server.py
+++ b/travis/test_server.py
@@ -136,10 +136,6 @@ def get_addons_path(travis_dependencies_dir, travis_build_dir, server_path):
     return addons_path
 
 
-def get_server_script(odoo_version):
-    return 'odoo-bin' if float(odoo_version) >= 10 else 'openerp-server'
-
-
 def get_addons_to_check(travis_build_dir, odoo_include, odoo_exclude):
     """
     Get the list of modules that need to be installed
@@ -183,7 +179,7 @@ def get_test_dependencies(addons_path, addons_list):
                 set(addons_list))
 
 
-def setup_server(db, odoo_unittest, tested_addons, server_path, script_name,
+def setup_server(db, odoo_unittest, tested_addons, server_path,
                  addons_path, install_options, preinstall_modules=None,
                  unbuffer=True, server_options=None):
     """
@@ -210,7 +206,7 @@ def setup_server(db, odoo_unittest, tested_addons, server_path, script_name,
     else:
         # unbuffer keeps output colors
         cmd_odoo = ["unbuffer"] if unbuffer else []
-        cmd_odoo += ["%s/%s" % (server_path, script_name),
+        cmd_odoo += ["%s/openerp-server" % server_path,
                      "-d", db,
                      "--log-level=info",
                      "--stop-after-init",
@@ -302,7 +298,6 @@ def main(argv=None):
             test_loghandler = 'openerp.tools.yaml_import:DEBUG'
     odoo_full = os.environ.get("ODOO_REPO", "odoo/odoo")
     server_path = get_server_path(odoo_full, odoo_version, travis_home)
-    script_name = get_server_script(odoo_version)
     addons_path = get_addons_path(travis_dependencies_dir,
                                   travis_build_dir,
                                   server_path)
@@ -331,14 +326,14 @@ def main(argv=None):
         os.environ.get('TRAVIS_BUILD_DIR'))))
     print("Modules to preinstall: %s" % preinstall_modules)
     setup_server(dbtemplate, odoo_unittest, tested_addons, server_path,
-                 script_name, addons_path, install_options, preinstall_modules,
-                 unbuffer, server_options)
+                 addons_path, install_options, preinstall_modules, unbuffer,
+                 server_options)
 
     # Running tests
     database = "openerp_test"
 
     cmd_odoo_test = ["coverage", "run",
-                     "%s/%s" % (server_path, script_name),
+                     "%s/openerp-server" % server_path,
                      "-d", database,
                      "--stop-after-init",
                      "--log-level", test_loglevel,
@@ -350,12 +345,12 @@ def main(argv=None):
 
     if odoo_unittest:
         to_test_list = tested_addons_list
-        cmd_odoo_install = [
-            "%s/%s" % (server_path, script_name),
-            "-d", database,
-            "--stop-after-init",
-            "--log-level=warn",
-        ] + install_options + ["--init", None] + server_options
+        cmd_odoo_install = ["%s/openerp-server" % server_path,
+                            "-d", database,
+                            "--stop-after-init",
+                            "--log-level=warn",
+                            ] + install_options + ["--init", None] +\
+                            server_options
         commands = ((cmd_odoo_install, False),
                     (cmd_odoo_test, True),
                     )
--- a/travis/test_server.py
+++ b/travis/test_server.py
@@ -298,6 +298,13 @@ def main(argv=None):
             test_loghandler = 'openerp.tools.yaml_import:DEBUG'
     odoo_full = os.environ.get("ODOO_REPO", "odoo/odoo")
     server_path = get_server_path(odoo_full, odoo_version, travis_home)
+    # Normalize on v10 odoo-bin command
+    try:
+        shutil.copy(
+            os.path.join(server_path, 'openerp-server'),
+            os.path.join(server_path, 'odoo-bin'))
+    except IOError:
+        pass
     addons_path = get_addons_path(travis_dependencies_dir,
                                   travis_build_dir,
                                   server_path)
@@ -333,7 +340,7 @@ def main(argv=None):
     database = "openerp_test"
 
     cmd_odoo_test = ["coverage", "run",
-                     "%s/openerp-server" % server_path,
+                     "%s/odoo-bin" % server_path,
                      "-d", database,
                      "--stop-after-init",
                      "--log-level", test_loglevel,
@@ -345,12 +352,12 @@ def main(argv=None):
 
     if odoo_unittest:
         to_test_list = tested_addons_list
-        cmd_odoo_install = ["%s/openerp-server" % server_path,
-                            "-d", database,
-                            "--stop-after-init",
-                            "--log-level=warn",
-                            ] + install_options + ["--init", None] +\
-                            server_options
+        cmd_odoo_install = [
+            "%s/odoo-bin" % server_path,
+            "-d", database,
+            "--stop-after-init",
+            "--log-level=warn",
+        ] + install_options + ["--init", None] + server_options
         commands = ((cmd_odoo_install, False),
                     (cmd_odoo_test, True),
                     )


pedrobaeza:
Why this change? Now it's working perfectly.

blaggacao:
@pedrobaeza Do you mean https://github.com/OCA/maintainer-quality-tools/issues/398 is accidentially still open?

travis/test_server.py
```diff
@@ -302,7 +298,12 @@ def main(argv=None):
             test_loghandler = 'openerp.tools.yaml_import:DEBUG'
     odoo_full = os.environ.get("ODOO_REPO", "odoo/odoo")
     server_path = get_server_path(odoo_full, odoo_version, travis_home)
-    script_name = get_server_script(odoo_version)
+    # Normalize on v10 odoo-bin command
+    try:
+      copy(os.path.join(server_path, 'openerp-server')),
```
moylop260: Sorry, my https://github.com/OCA/maintainer-quality-tools/issues/398#issuecomment-267404414 was just a  example of the idea. It shouldn't considered real code.

pedrobaeza:
"master" is not a valid branch name, as there's a lot of stuff derived from the branch name, so although you change this, you will find a lot of troubles. Why don't you simply use the current conventions?

blaggacao:
@pedrobaeza I was just reading over @moylop260 's issue and created a sample implementation of his idea. I guess the only one who can tell us is @moylop260 , actually. I have no personal interest in this whatsoever. It just seemed so simple and quick to me, that I decided give @moylop260 a hand and help him with the discussion by providing a substrate upon which can be discussed.

travis/test_server.py
blaggacao: No worries, I liked your idea... Doing this was a 50 seconds task... Those you should do immediately 💃 😄

pedrobaeza:
OK, let's see if he wants to continue with the idea or not.

moylop260:
@pedrobaeza 
if we want test a feature from `master` or `saas-*` valid branches from official repository(1) we can not use them since merge of https://github.com/OCA/maintainer-quality-tools/pull/364

This PR add accidentally a new lock to use just branches name with version.
Now the original author is agree with the change of this PR (2)

(1) https://github.com/odoo/odoo/branches/active
(2) https://github.com/OCA/maintainer-quality-tools/issues/398#issuecomment-267426594

pedrobaeza:
OK, but you know (as you have developed it), that there are other values inferred from the branch name for configuring things in MQT, so this is only one minor problem I think.

blaggacao:
From the incomplete list, its almost a case about deprecating "version" in favor of "branch"...
Unproblematic: https://github.com/OCA/maintainer-quality-tools/blob/master/travis/test_server.py#L39
Doesn;t even use: https://github.com/OCA/maintainer-quality-tools/blob/master/travis/test_server.py#L240
https://github.com/OCA/maintainer-quality-tools/blob/master/travis/test_server.py#L110
Easy: https://github.com/OCA/maintainer-quality-tools/blob/master/travis/clone_oca_dependencies#L48
Nice: https://github.com/OCA/maintainer-quality-tools/blob/a6e6aea1f527c74c2d905bb595d1216a791c4bb4/travis/test_pylint#L33

pedrobaeza:
Thanks for the initial list, David. That's what I refer to when talking about other issues, but if you handle all, then no problem.

blaggacao:
OK, let's do that... Meanwhile you could probably help with 
https://github.com/OCA/maintainer-quality-tools/pull/410 and https://github.com/OCA/maintainer-quality-tools/pull/408

blaggacao:
Closed in favor of #411

travis/test_server.py
```diff
@@ -136,10 +136,6 @@ def get_addons_path(travis_dependencies_dir, travis_build_dir, server_path):
     return addons_path
 
 
-def get_server_script(odoo_version):
-    return 'odoo-bin' if float(odoo_version) >= 10 else 'openerp-server'
```
dreispt: This change doesn't look good: you need this to support Odoo 10.

travis/test_server.py
```diff
@@ -350,12 +351,12 @@ def main(argv=None):
...
-            "-d", database,
-            "--stop-after-init",
-            "--log-level=warn",
-        ] + install_options + ["--init", None] + server_options
+        cmd_odoo_install = ["%s/openerp-server" % server_path,
+                            "-d", database,
+                            "--stop-after-init",
+                            "--log-level=warn",
+                            ] + install_options + ["--init", None] +\
+                            server_options
```
dreispt: Please don't change code formatting unless it is really necessary.

dreispt:
Instead of just reverting a previous PR, I expect a change to just fix the issue described in #398

travis/test_server.py
blaggacao: The idea is to accomplish it by the proposed file copy _instead_ of using this method, which introduced the problems described in #398 in the first place. HI doubt having two alternative implementations of the same thing is what we want.

travis/test_server.py
blaggacao: Actually I haven't introduced them. This is what I introduced besides of reverting the previous commit: https://github.com/OCA/maintainer-quality-tools/pull/409/commits/9696b3aba595c810f7a26fa08bfd6a9b57655d65
Remember that @StefanRijnhart has agreed to revert it. At least that's what his [comment](https://github.com/OCA/maintainer-quality-tools/issues/398#issuecomment-267426594) suggests. I see now, how complicated distributed communication and coordination is 😄. I'm realizing this, while trying to keep out of the line of fire somehow...

blaggacao:
@moylop260 Is this a valid implementation of your idea or do you prefer not to follow up on this? ("Formally approvable by you?") If it's not we should probably know and stop spending time on this... Although it's always a great pleasure working together in an in depth review with @dreispt 😄

moylop260:
@blaggacao 
I told you before... this idea is fine to me thanks for help us
Now you need a travis green for my `formally approvable`

moylop260:
> @dreispt  Instead of just reverting a previous PR...

The fix require the revert because we don't need the new variable added and we have 2 commits here:
 1. revert commit
 2. fix commit

Maybe It to avoid noise with the diff you could review just the last commit.

blaggacao:
@moylop260 Thanks for your support. And sorry, I couldn't tell, whether you were pleased or not. Espescially from https://github.com/OCA/maintainer-quality-tools/pull/409#discussion_r95270768 
So, being reassured, I'll proceed with reviewing travis now...

moylop260:
The https://github.com/OCA/maintainer-quality-tools/pull/409#discussion_r95270768 means:
 - `This line of code will be travis red. Sorry If you think that is a real line of code working`
But the idea is the same

blaggacao:
@moylop260 Oh well and I almost thought you were against it for some reason... 😄 Thanks for clarifying.

blaggacao:
@moylop260 OK, now most of the travis tests are clean. I'm not sure about the remaining:

- https://travis-ci.org/OCA/maintainer-quality-tools/jobs/191419989 seems to be expected to fail, as there is something about the broken_module, which fails, but overall summaries show green, yet travis itself fails.
- https://travis-ci.org/OCA/maintainer-quality-tools/jobs/191420041 fails seemingly on pre-commit script. Is that possibly due to this PR?
- https://travis-ci.org/OCA/maintainer-quality-tools/jobs/191420042 same as above...

travis/test_server.py
```diff
@@ -302,7 +298,12 @@ def main(argv=None):
             test_loghandler = 'openerp.tools.yaml_import:DEBUG'
     odoo_full = os.environ.get("ODOO_REPO", "odoo/odoo")
     server_path = get_server_path(odoo_full, odoo_version, travis_home)
-    script_name = get_server_script(odoo_version)
+    # Normalize on v10 odoo-bin command
+    try:
+        shutil.copy(os.path.join(server_path, 'openerp-server'),
+            os.path.join(server_path, 'odoo-bin'))
```
StefanRijnhart: You copy openerp-server to odoo-bin, but then you kept the references to openerp-server everywhere.

travis/test_server.py
StefanRijnhart: Indeed this is a revert of my initial solution, followed by a more elegant and pythonic solution in the second commit.

travis/test_server.py
blaggacao: Thanks, solved. Sorry for this oversight. Need to get accustomed to the multiple challenges of contributing while not losing focus on the code.. 😄

blaggacao: (force pushed)

blaggacao:
@moylop260 done, thanks for copying them... 😄

moylop260:
Please, Could you rebase the lint commit with previous one?

blaggacao:
You're welcome. 😃

moylop260:
@pedrobaeza @dreispt 
Could you merge it, please?

travis/test_server.py
```diff
@@ -302,7 +298,13 @@ def main(argv=None):
             test_loghandler = 'openerp.tools.yaml_import:DEBUG'
     odoo_full = os.environ.get("ODOO_REPO", "odoo/odoo")
     server_path = get_server_path(odoo_full, odoo_version, travis_home)
-    script_name = get_server_script(odoo_version)
+    # Normalize on v10 odoo-bin command
+    try:
+        shutil.copy(
+            os.path.join(server_path, 'openerp-server'),
+            os.path.join(server_path, 'odoo-bin'))
```
dreispt: I have a concern about this copy operation: it is intrusive, in the sense that it makes modification on the code being used for the tests. This is will not only be unexpected, it will also be an issue when MQT will be used to run tests on our local dev machine. It changes a "stateless" operation into a "stateful" one, for no benefit at all.

Again, why is this change necessary?

travis/test_server.py
dreispt: I suggest fixing this function by handling the special case `odoo_version=='master'`

travis/test_server.py
moylop260: Could you see the related issue https://github.com/OCA/maintainer-quality-tools/issues/398?

moylop260: ...and comment about https://github.com/OCA/maintainer-quality-tools/pull/409#issuecomment-271448719

blaggacao: @dreispt I understand your concerns. However, in a certain way at least for the local machine case, it is most probably necessary to construct a dedicated testbed, which is initiated for example by a rsync command. It could be a problem if you bind mount read only indeed. But, if we are operating on read only filesystems, the whole `travis_install_nightly` will not work...
I think the solution is a "testbed", a word which i already used several times now:
A dedicated folder with predefined permissions and for single use. So stateful changes, do not matter. BTW, there are other stateful changes relevant in testing: Think you want to inhabilitate a testing of a core module in order to replace it with a custom test via `sed` command. So I guess the concept of (one way) testbed is pretty appropriate and stateful changes are not of concern.

blaggacao: @dreispt Could the "testbed argument" solve your concerns and help to release the blocking objections?

dreispt: In a dev machine the test bed is the WIP code, and that may include a patched Odoo repo.
What is the problem with just fixing the bug in the function deciding the Odoo executable name?

blaggacao: Because the original function innecessarily increases the signature of another function and thereby makes the code more complex, harder to read and harder to maintain (workaround approach). This solution takes the aligns things and keeps them parallel (normalization approach).

blaggacao: If you men WIP code = working directory, then I need to correct you. The test bed is a copied (or rsynced) special folder structure in the most common case here... From this copy, nothing is fed back, never. You could call it prep bed, for "preparation". This has almost the word "stateful" inherent. If for any improbable chance the working directory folder structure happens to coincide with the test bed folder structure (in case of adding enterprise it cant, for example), we should rather force the use of a testbed, there...

blaggacao: BTW `get_modules_changed` method is stateful, when it fetches, and as the <src> part of the refspec is undefined there could just pop up a new branch in your repo unexpectedly.

PR #384, opened 2016-11-14
[REF] .travis.yml: Enable test to install just a module different

moylop260:

PR #410, opened 2017-01-10
[FIX] oversight from #392

blaggacao:
#392 introduced the ability to populate build and dependency dirs with several addons. Yet it oversaw to align these two references to travis_build_dir. /cc @dreispt @moylop260 @pedrobaeza


moylop260:
Sorry I didn't understand this change.
Could you add a comment with the directory structure that you are proposing?

blaggacao:
https://github.com/OCA/maintainer-quality-tools/pull/392 from @dreispt enables the following directory structure, this commit is a fixup on #392 only, it does not introduce new behaviour.
```
home/dependencies/folder1/addon1
home/dependencies/folder1/addon2
home/dependencies/folder123/addon3
home/dependencies/folder123/addon4
home/build/folder1/addon1
home/build/folder1/addon2
home/build/folder123/addon3
home/build/folder123/addon4
```

blaggacao:
@jcdrubay you should have run on this as well in your investigation after #392 ?

dreispt:
Before #392, MQT assumed that the "build_dir" was an addons directory, as is the case for OCA tests:

    home/build/addon1
    home/build/addon2

With #392 MQT can also handle the case where the "build_dir" contains several addons directories.

    home/build/folder1/addon1
    home/build/folder1/addon2
    home/build/folder2/addon3

I'm missing what needs fixing.

blaggacao:
@dreispt have you looked my proposed changes? Aren't they somehow obvious?

blaggacao:
@dreispt Let me know if you need any further information. I'm happy to provide everything necessary, that might help to get this merged quickly.

blaggacao:
Sorry, I'm terrible, but this is itching... 😄 And its a bug fix, if that merits different treatment. 🐝

dreispt:
@blaggacao I must be honest in that I find quite difficult to follow your explanations.

If I understand correctly, my fix correctly populates the addons list, but fails to populate the list of modules to preinstall. Is that it?

Your code looks good, but I propose that instead of it, you change the [get_modules](https://github.com/OCA/maintainer-quality-tools/blob/master/travis/getaddons.py#L49) function to also work in the case of a directory of addon directories (similarly to the function you added here).

What do you think?

blaggacao:
@dreispt I guess, we are on the same page now. Thanks for taking the time to understand my line of thought. I'll have a look if get_modules is out of any risk, then this would be even better, of course.

blaggacao:
Ok, I come back on this one. From source:
```
def get_modules(path):

    # Avoid empty basename when path ends with slash
    if not os.path.basename(path):
        path = os.path.dirname(path)

    res = []
    if os.path.isdir(path):
        res = [x for x in os.listdir(path)
               if is_installable_module(os.path.join(path, x))]
    return res


def is_addons(path):
    res = get_modules(path) != []
    return res


def get_addons(path):
    if not os.path.exists(path):
        return []
    if is_addons(path):
        res = [path]
    else:
        res = [os.path.join(path, x)
               for x in os.listdir(path)
               if is_addons(os.path.join(path, x))]
    return res
```
Here it goes `is_addons` checks if the current folder is an addons folder, that is if its direct child folders are modules. It tests so via `get_modules`. I'd need to refactor this part of the code, but I guess it will be better design, afterwards...

blaggacao:
@dreispt https://github.com/OCA/maintainer-quality-tools/pull/410/commits/34ced5c415f58228f7f5841a412a86d0129fe4e1 is a more generalized refactoring, but due to the reaons mentioned in my above comment, I could not do a minimal change in order to realize this approach. However, I think that this might be even a better design. We still have the original functions, which do have it's very own valid use case, but we added a recursive search to handle more advanced folder structures genuinely.

blaggacao:
I guess now its either a squash of the whole branch or a cherry pick of the commits before https://github.com/OCA/maintainer-quality-tools/pull/410/commits/34ced5c415f58228f7f5841a412a86d0129fe4e1

dreispt:
@blaggacao Let's keep complexity low: refactor the existing `get_addons` and `get_modules` functions, instead of adding new ones.
Also, don't do edits that just change code style, unless they fix lint warnings.

blaggacao:
@dreispt I'm afraid, your first recommendation is not possible without introducing regressions... see https://github.com/OCA/maintainer-quality-tools/pull/410#issuecomment-272296922 for details. 
Actually I don't see any code style changes [here](https://github.com/OCA/maintainer-quality-tools/pull/410/files), where your recommendation would apply. (the import line would fail because it's too long)

---------------------------------

Guys, don't overly mistrusting. All of you! 🐱 It holds the risk of being wrong. And it greatly improves the cost of contributing at no additional benefit. It might happen, that I oversee small things, and I am very grateful if you show them to me. However, I feel regularly embarrassed in these discussion these days, as I feel not in the position nor does it give me pleasure to show someone that he is wrong and it is perceivably happening quite a lot... No offence, please, just a perceived impediment. I'm on your side, after all.

---------------------------------

So I'll leave this PR how it is unless there is anything more I can do to improve it. Are you ok with this?

blaggacao:
seized (closes) #384 on the go...

blaggacao:
**T r a v i s i s**
🍏 🍏 🍏

* [x] Tests
* [x] Travis
* [x] Rebase

PR #411, opened 2017-01-10
Fix master version

blaggacao:
# replaces https://github.com/OCA/maintainer-quality-tools/pull/409 
Copied from https://github.com/OCA/maintainer-quality-tools/pull/409#issuecomment-271448719:

> if we want test a feature from master or saas-* valid branches from official repository(1) we can not use them since merge of #364
> 
> This PR add accidentally a new lock to use just branches name with valid version.
> Now the original author is agree with the change of this PR (2)
> 
> (1) https://github.com/odoo/odoo/branches/active
> (2) #398 (comment)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,30 +23,30 @@ virtualenv:
 env:
 
   global:
-  - VERSION="9.0" TESTS="1" LINT_CHECK="0"
+  - BRANCH="9.0" TESTS="1" LINT_CHECK="0"
 
   matrix:
   - INSTANCE_ALIVE="1" INCLUDE="broken_module" SERVER_EXPECTED_ERRORS="2"  # test errors are detected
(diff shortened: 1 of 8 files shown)


blaggacao:
@moylop260 If this is what you were thinking about, then there is only one very little change in pylint_odoo necessary `--valid_versions` > `--valid_branches`. I haven't looked into pylint_odoo, so I have no Idea if this is in the best interest of it's creator. This PR, however is an attempt to give the `version` concept a more concise reference by renaming it to `branch`.

blaggacao:
I made it backward compatible (https://github.com/OCA/maintainer-quality-tools/pull/411/files#diff-17b9133d6cebf66784dc392176486372R35). I did this upon how I understood pedro's comment in the other commit. So I won't argue anything, except that either here or in #409 is a fix for the problem exposed in #398 .

moylop260:
If you have a `"WARNING: env variable VERSION is deprecated, use BRANCH instead."` a massive change to `.travis.yml` is required yet.

We don't need a renamed of variable here to fix it.
We can work with `VERSION="saas-4"` or `VERSION="master"` like as https://github.com/OCA/maintainer-quality-tools/pull/409

PR #416, opened 2017-01-11
[FIX] full api compliance with odoo command options

### LINKED ISSUE #414: [Issue] createdb does not use server options (host, user, pw)
**What is it**
the `createdb` commands are not fed the (potential) db, user and database server options, which are received by the odoo command. Instead they solely obey to standard pg env variables (`PGHOST`, `PGUSER`, `PGPASSWORD`).

**Why is it boring**
You might spend some time finding out this somewhat unintuitive behaviour.

**Why is it (almost) blocking (in some cases)**
In some CIs (eg gitlab-ci) you can define the used database as "service", internally it just downloads a docker images and binds it to the runner container under it's name. So if you define service `postgres` it will be available under the hostname `postgres` to your services. Because of the above said, you then define unsuspectingly `PGHOST=postgres`. Results, that the whole environment is shared to postgres image, so it's run command and init script will fail, as it is supposed to run against `localhost`, not `postgres`. 

**Rationale**
The argument here is that in a given environment the main consumer of `PG*` variables should be postgres itself, MQT should not conflictingly consume those variables, but rather stick with the more scoped ones contained in `SERVER_OPTIONS`. This is a typical design problem of the config by environment variable pattern. I suggest preparing a fix which makes MQT give precedence to `SERVER_OPTIONS` over `PG*` variables, if set. 

@moylop260  would you accept such PR?

**I can guess the questions...**
- Yes, it can be worked around by sub shelling `(export PGHOST=postgres; export PGUSER=Testuser; export PGPASSWORD=Testpassword; travis_run_tests)`
- No, although a necessary solution, this is not sufficient, as we do not want to twist people's minds for no particular reason, except for own comfort. We should be nice with the user, shouldn't we?

blaggacao:
closes #414 
Signed-off-by: David Arnold <dar@devco.co>

### DIFF
--- a/travis/test_server.py
+++ b/travis/test_server.py
@@ -192,15 +192,16 @@ def prep_pg_environ(server_options):
     :param server_options: (list) Add these flags to the Odoo server init
     """
     parser = OptionParser()
-    parser.add_option("-r", "--db_user", dest="db_user", my_default=False)
-    parser.add_option("-w", "--db_password", dest="db_password", my_default=False)
-    parser.add_option("--pg_path", dest="pg_path", my_default=False)
-    parser.add_option("--db_host", dest="db_host", my_default=False)
-    parser.add_option("--db_port", dest="db_port", my_default=False)
+    parser.add_option("-r", "--db_user", dest="db_user", default=False)
+    parser.add_option(
+        "-w", "--db_password", dest="db_password", default=False)
+    parser.add_option("--pg_path", dest="pg_path", default=False)
+    parser.add_option("--db_host", dest="db_host", default=False)
+    parser.add_option("--db_port", dest="db_port", default=False)
     (options, args) = parser.parse_args(server_options)
     pg_env = os.environ.copy()
     if options.db_host:
-        pg_env['PGHOST'] = options.db_host 
+        pg_env['PGHOST'] = options.db_host
     if options.db_user:
         pg_env['PGUSER'] = options.db_user
     if options.db_password:
@@ -233,8 +234,8 @@ def setup_server(db, odoo_unittest, tested_addons, server_path, script_name,
         server_options = []
     print("\nCreating instance:")
     try:
-        pg_env = os.environ.copy()
-        subprocess.check_call(["createdb", db], 
+        subprocess.check_call(
+            ["createdb", db],
             env=prep_pg_environ(server_options))
     except subprocess.CalledProcessError:
         print("Using previous openerp_template database.")
@@ -400,7 +401,7 @@ def main(argv=None):
         db_odoo_created = False
         try:
             db_odoo_created = subprocess.call(
-                ["createdb", "-T", dbtemplate, database], 
+                ["createdb", "-T", dbtemplate, database],
                 env=prep_pg_environ(server_options))
             copy_attachments(dbtemplate, database, data_dir)
         except subprocess.CalledProcessError:
--- a/travis/test_server.py
+++ b/travis/test_server.py
@@ -199,25 +199,25 @@ def get_local_env(server_options):
     parser.add_option("--db_host", dest="db_host", default=False)
     parser.add_option("--db_port", dest="db_port", default=False)
     (options, args) = parser.parse_args(server_options)
-    pg_env = os.environ.copy()
+    env = os.environ.copy()
     if options.db_host:
-        pg_env['PGHOST'] = options.db_host
+        env['PGHOST'] = options.db_host
     if options.db_user:
-        pg_env['PGUSER'] = options.db_user
+        env['PGUSER'] = options.db_user
     if options.db_password:
-        pg_env['PGPASSWORD'] = options.db_password
+        env['PGPASSWORD'] = options.db_password
     if options.db_port:
-        pg_env['PGPORT'] = options.db_port
+        env['PGPORT'] = options.db_port
     # This is not supported to be passed to PG* variables, 
     # so it should break entirely as no api compliance can be achieved easily.
     if options.pg_path:
         raise "Option --pg_path is not supported in MQT."
-    return pg_env
+    return env
 
 
 def setup_server(db, odoo_unittest, tested_addons, server_path, script_name,
-                 addons_path, install_options, local_env,
-                 preinstall_modules=None, unbuffer=True, server_options=None):
+                 addons_path, install_options, preinstall_modules=None,
+                 unbuffer=True, server_options=None):
     """
     Setup the base module before running the tests
     if the database template exists then will be used.
@@ -255,7 +255,7 @@ def setup_server(db, odoo_unittest, tested_addons, server_path, script_name,
     return 0
 
 
-def run_from_env_var(env_name_startswith, environ, local_env):
+def run_from_env_var(env_name_startswith, environ):
     '''Method to run a script defined from a environment variable
     :param env_name_startswith: String with name of first letter of
                                 environment variable to find.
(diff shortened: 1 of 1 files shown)


travis/test_server.py
```diff
@@ -204,7 +233,9 @@ def setup_server(db, odoo_unittest, tested_addons, server_path, script_name,
         server_options = []
     print("\nCreating instance:")
     try:
-        subprocess.check_call(["createdb", db])
```
moylop260: What about `subprocess.check_call(["createdb", db], env=os.environ)`?
To send all environment variables avoiding a wrapper `prep_pg_environ` method.

blaggacao: Can it be, that you have misunderstood #414? I'm a bit puzzled about your idea...

blaggacao:
**T r a v i s  i s**
🍏 🍏 🍏

moylop260:
Sorry If I'm not clear with my https://github.com/OCA/maintainer-quality-tools/issues/414#issuecomment-271787391

That means we need propagate the original `os.environ` for `pipe` command methods.
`createdb` use environment variables to get postgresql connection
`odoo` use environment variables to get postgresql connection too. (You could confirm it using https://hub.docker.com/_/odoo/ directly where PG params is not used just PG environments variables)

Then we should use [PG environment variables ](https://www.postgresql.org/docs/8.4/static/libpq-envars.html) instead of parameters.
Because if you need special parameters for `createdb` command then you will need special parameters for `odoo` and you will need create a new `prep_pg_environ_odoo` method for this case.

If you use 100% environment variables we don't need a wrapper map of parameters of PG for odoo and psql and other one.

blaggacao:
Yes, I do understand what you are saying: 

> You are saying that we need to propagate `os.environ` for `pipe` commands.

My response to that was: the environment is already present for `pipe` commands.

> odoo use environment variables to get postgresql connection too

I already looked at that, because I was wondering, too. The clues is here >(https://github.com/odoo/docker/blob/master/10.0/entrypoint.sh#L5-L24). It doesn't use `PG*` environment variables, though, but a different set of docker specific env variables. (I personally do not recommend using the official odoo docker in any way.)

> Then we should use PG environment variables instead of parameters.
Because if you need special parameters for createdb command then you will need special parameters for odoo and you will need create a new prep_pg_environ_odoo method for this case.

I cannot follow this argument (I think I don't understand it).

> If you use 100% environment variables we don't need a wrapper map of parameters of PG for odoo and psql and other command that requires this values.

If we do that, we give away api-compliance with the standard odoo command. I'd rather prefer to start a discussion upstream, so that odoo includes config by environment variables into it's configuration layout. Something like `ODOO_*`.

moylop260:
> I personally do not recommend using the official odoo docker in any way.

I don't use official odoo docker.
This is just to show you that the parameters is not required if you use PG* environment variables easiest 😄 

> I cannot follow this argument (I think I don't understand it).

Ok, I will send you a video about this matter.

blaggacao:
>This is just to show you that the parameters is not required if you use PG* environment variables easiest 😄

But odoo **does not** ues `PG*` environment variables, nor does odoo docker. Unless I really badly missed anything?

moylop260:
That is my point here: odoo does USE PG* environment variables

I mean
[asciicast](https://asciinema.org/a/99305?t=0:14)

blaggacao:
Stupid me of course it does, as psycopg2 does use the postgres binaries... OK, thanks a lot!

Then this commit just helps with api compliance (the server options). People might still use it and for some it might even be the preferred method, as it writes the config file for them... It's a very bad design decision from Odoo, I know, but for the time being, I guess we should comply and rather bring the discussion upstream to work with a different config mechanism.

So to resume, what this commit does: If no `PG*` environment variable is set, then it falls back to the ones specified in the server options. Now, the question is is this api compliant? Doesn't odoo prefer its own server option over the `PG*` variables? If so I would need to change this PR...

moylop260:
IMHO this PR should add all items (including testing) to work with default `PG*` environment variables because currently is not compatible.

Other matter (and other PR) is use your special wrapper method (I'm not agree with this method, but maybe other reviewers are agree, but reduce the quantity of changes could help to merge this one)

blaggacao:
> IMHO this PR should add all items (including testing) to work with default PG* environment variables because currently is not compatible.

You mean under the hood extract the server options and set `PG*` options globally (for the scope of `test_server`)? Must be scoped as `PG*` on the global environment can be undesirable as they can be shared as a meta-environment to side kick containers where they can harm...

> Other matter (and other PR) is use your special wrapper method (I'm not agree with this method, but maybe other reviewers are agree, but reduce the quantity of changes could help to merge this one)

OK, this was designed mainly to solve https://github.com/OCA/maintainer-quality-tools/pull/407/files#diff-04c6e90faac2675aa89e2176d2eec7d8R112 as line 112 will fail, when we have the above mentioned meta-environment shared to side kick containers.

Headsup:
A `postgres` side kick would use those `PG*`  variables as well, but it's init script is supposing that the server listens on localhost, so if we define `PGHOST != localhost`, the init script will fail, and thus line 112 from the above docs would fail, as well.

blaggacao:
@moylop260 I think I got your point while refactoring, you would like to have "homogeneousity" throughout the code. I now generalized the wrapper method, so it can be also used for other use cases. It's definitely better now, and it is by default passed to all `subprocess` calls (note that I had missed one relevant for `dropdb` in my previous commits, fixed).

We cannot drop the requirement that this environment must be local to `test_server`, though as argumented above. So I guess now this is the best out of my initial idea to solve my problem and your recommendations.

blaggacao:
@moylop260 The second commit, now implements an alternative using a global variable for the script-local, but otherwise global environment. This seems the most elegant solution to me. An argument for using global variable is directly in the commit message.

blaggacao:
> If you use 100% environment variables we don't need a wrapper map of parameters of PG for odoo and psql and other command that requires this values.

This would break then the use case where there is a shared meta-environment with sidekick containers.

moylop260:
If you wish create a feature to add a wrapper for use the same parameters of PG you should split this PR in 2.

I'm agree for use of `env` parameter
But the other one I need reviewers feedback.

Then we can merge first just the env parameter case for now.

blaggacao:
OK, but this would then not solve #414 which is my underlying motivation and drive for doing this... So where is the "tit for tat" here?

@lasley help me stand a little bit my ground, as you said... this is feeling to get "Haarspalterei" (look it up, it's a nice german saying! 😃 ).

travis/test_server.py
```diff
@@ -9,6 +9,7 @@
 import sys
 from getaddons import get_addons, get_modules, is_installable_module
 from travis_helpers import success_msg, fail_msg
+from optparse import OptionParser
```
lasley: This is deprecated for `argparse`

lasley:
TBH I'm somewhat confused on the scope of this PR, so I have been staying away from it.

I'd be happy to provide more opinion here, but can you please make the title and description of this issue more meaningful? While the argument can be made that I can simply read the code, a reviewer needs to be able to understand the full scope of your intent before being able to make a determination on whether the code fits that scope. I personally do not know what the scope of "full api compliance with odoo command options" is supposed to encompass & the issue referenced in the description is a bit dense in terms of description for my ADD.

moylop260:
I understood from issue related that you just want propagate the `PG*` environment variables to popen commands :+1:
But create a wrapper to send them from parameters is too much for the MQT scripts 👎 
If a new parameters is added in future versions of PG then we need maintain this feature, but Is really need?
What about use directly the [official PG environment variables](https://www.postgresql.org/docs/8.4/static/libpq-envars.html) and avoid maintain cruft code?

lasley:
> But create a wrapper to send them from parameters is too much for the MQT scripts

Yeah the wrapper is kind of the part where I got lost in terms of relating the scope and all that. 

I'm definitely all for the env vars though & think it's a great improvement.

> What about use directly the official PG environment variables

👍  - looks like this PR conforms to the correct names too, it's just a matter of passing them through.

PR #419, opened 2017-01-14
[FIX] test based on wrong use of zip()

blaggacao:


moylop260:
A rebase is required here in order to get just the last commit and fix conflicts

blaggacao:
OK, no problem, now that's in master.. 😄

moylop260:
Fast-tracking since that is a small fix from testing

PR #421, opened 2017-01-15
[FIX] for local testing and fix wrong variable attribution

blaggacao:
**What is it**
It's a bug fix. So there was a bug beforehand.

**Why is it**
- `repo_dir = os.environ.get("TRAVIS_BUILD_DIR", ".")` this is ok. `TRAVIS_BUILD_DIR` usually coincides with '.' (seen from the `.travis.yaml`, see `coverage run --append ./travis/self_tests`)
- But already `to_preinstall = get_test_dependencies(repo_dir , addons_list)` fails as `get_test_dependencies` is not recursive, it needs the actual addons dir with is at `./tests/repo_dir`.
- And so forth ...
The only exception is
- `getaddons.get_modules_changed(repo_dir)` which works on the folder which contains the `.git` folder.

With these changes, I was able to seamlessly run the test script locall.

image

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,59 @@
-sample_files/.gitignore
+# Byte-compiled / optimized / DLL files
+__pycache__/
+*.py[cod]
+
+# C extensions
+*.so
+
+# Distribution / packaging
+.Python
+env/
+bin/
+build/
+develop-eggs/
+dist/
+eggs/
+lib/
+lib64/
+parts/
+sdist/
+var/
+*.egg-info/
+.installed.cfg
+*.egg
+
+# Installer logs
+pip-log.txt
+pip-delete-this-directory.txt
+
+# Unit test / coverage reports
+htmlcov/
+.tox/
+.coverage
+.cache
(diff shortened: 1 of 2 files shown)


blaggacao:
On the other hand, the intent to link a gitignore to another gitignore did not work in my case...

blaggacao:
@moylop260 Could we push this true, as it would make my fully test compliant work on MQT a lot easier, faster, less error prone and less noisy?

.gitignore
```diff
@@ -1 +1,59 @@
-sample_files/.gitignore
\ No newline at end of file
+# Byte-compiled / optimized / DLL files
```
moylop260: Is this change required?

blaggacao: Yep, this linking does not work on all (my) platform(s). Greetings!

.gitignore
```diff
@@ -1 +1,59 @@
-sample_files/.gitignore
```
blaggacao: I mean this linking here...

PR #425, opened 2017-01-15
On TravisCI stop alive process to get green builds

dreispt:


dreispt:
@moylop260 Can you check if this fix suits you?

moylop260:
This is a good solution to me.
Thanks

Just we need test this parameter from runbot.
Could you create a dummy PR to any green oca project to use this branch?

dreispt:
> Just we need test this parameter from runbot.

It already is.

> Could you create a dummy PR to any green oca project to use this branch?

I don't see the point: the issue is only with MQT self tests.

dreispt:
This PR is just about fixing the self-tests to the INSTANCE_ALIVE feature, so that we can get green CI builds again.

PR #426, opened 2017-01-15
Fix error when VERSION is master. Fixes #398

### LINKED ISSUE #398: ValueError: could not convert string to float: master

dreispt:
This alternative to the fix provided by #409. IMO this is simpler and not intrusive on the code used to run the tests.

### DIFF
--- a/travis/test_server.py
+++ b/travis/test_server.py
@@ -137,7 +137,11 @@ def get_addons_path(travis_dependencies_dir, travis_build_dir, server_path):
 
 
 def get_server_script(odoo_version):
-    return 'odoo-bin' if float(odoo_version) >= 10 else 'openerp-server'
+    """Command to start Odoo, depending on the version"""
+    script = 'odoo-bin'
+    if odoo_version != 'master' and float(odoo_version) >= 10:
+        script = 'openerp-server'
+    return script
 
 
 def get_addons_to_check(travis_build_dir, odoo_include, odoo_exclude):


moylop260:
I prefer to use the concluded solution from https://github.com/OCA/maintainer-quality-tools/issues/398 by other contributors where you have a comment too

blaggacao:
Can we produce a conclusion, here? I'm not 100% convinced that it would be worthwhile to invest time and effort to get this sorted out. My suggestion is @dreispt to step back on this, and look how #411 behaves on production (like in opportunistic merging), we can put a big question mark on it, so if it needs to be reversed, then referencing this question mark, could provoke a rather swift fix on this, without the need to repeat this discussion, would that be ok? I'd like to push other topics, too, and being blocked on this consumes focus.

blaggacao:
@dreispt Can I stress again on moving forward on #411 please? I could not reliably follow the line of thought of what's the specific benefit of this blocking here?
Since when mere preferences becoming hurdle objections does foster collaboration?

dreispt:
I agree that this fix has strong limitations, so I'm closing this.
But I still disagree that renaming the Odoo startup script file is the proper solution.
A good solution could be to inspect if `odoo.py` exists, and if not use `openerp.py`.

Issue #173, opened 2015-04-01
[master] MQT: Status red


moylop260:
![descarga 4](https://cloud.githubusercontent.com/assets/6644187/6954321/97c09804-d88d-11e4-91b7-d6b2554b554f.png)

dreispt:
The MQT tests for 7.0 and 6.1 are failing - it seems that something is is now incompatible with those Odoo versions.

I believe it is caused by the [test dependency](https://github.com/OCA/maintainer-quality-tools/commit/b43cc487c02c9b927eb527d6b26727e00f0f0df7#diff-08357642cfa8d94d6869b876051508beR4) on `partner-contact` for v8.
It should be removed from the `test_repo`.

PR #428, opened 2017-01-25
[IMP] Support branches in the environment variable VERSION of .travis.yml 

JesusZapata:
Support branches in the environment variable VERSION of .travis.yml

Now
```
VERSION=master
ODOO_REPO=odoo/odoo
```
Expected
```
VERSION=pull/3
ODOO_REPO=odoo/odoo
```

In the enviroment variable `VERSION` can be used:
- [x] Branch : VERSION="master"
- [x] PR : VERSION="pull/10

Referring to issue #173

### DIFF
--- a/travis/self_tests
+++ b/travis/self_tests
@@ -82,7 +82,7 @@ if os.environ.get('LINT_CHECK', 0) == '1':
 getaddons.get_modules_changed(repo_dir)
 
 # Testing get_server_script
-assert get_server_script("/tmp") == 'odoo-bin'
+assert get_server_script("/tmp") == 'openerp-server'
 
 # Testing instance running
 def connection_test():
--- a/travis/self_tests
+++ b/travis/self_tests
@@ -10,6 +10,7 @@ import getaddons
 import travis_helpers
 from test_server import main as test_server_main
 from test_server import get_test_dependencies
+from test_server import get_server_script
 
 repo_dir = os.environ.get("TRAVIS_BUILD_DIR", "./tests/test_repo/")
 exclude = os.environ.get("EXCLUDE")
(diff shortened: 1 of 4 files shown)


travis/test_server.py
```diff
@@ -136,8 +137,13 @@ def get_addons_path(travis_dependencies_dir, travis_build_dir, server_path):
     return addons_path
 
 
-def get_server_script(odoo_version):
-    return 'odoo-bin' if float(odoo_version) >= 10 else 'openerp-server'
+def get_server_script(server_path):
+    try:
+        shutil.copy(os.path.join(server_path, 'openerp-server'),
+                    os.path.join(server_path, 'odoo-bin'))
+    except IOError:
```
moylop260: Add `# pragma: no cover` comment or add the correct test case from selftest

JesusZapata: @moylop260 
You can help me?
Explained as run self_test ?

travis/self_tests
```diff
@@ -80,6 +81,8 @@ if os.environ.get('LINT_CHECK', 0) == '1':
 # Testing git run from getaddons
 getaddons.get_modules_changed(repo_dir)
 
+# Testing get_server_script
+assert get_server_script("/tmp") == 'odoo-bin'
```
JesusZapata: @moylop260 
Is this the expected test?

JesusZapata:
@moylop260 
In this PR now is also the solution to this issue https://github.com/Vauxoo/maintainer-quality-tools/issues/178

JesusZapata:
@moylop260 
In this PR now is also the solution to this issue https://github.com/Vauxoo/maintainer-quality-tools/pull/181

travis/test_server.py
dreispt: It's best for the `get_server_script` to not have side effects.
You should instead check if `odoo-bin` exists, and fall back to `openep-server` if not.

JesusZapata: @moylop260 
What do you think about the previous comment?
I do not remember where PR agreed this change

moylop260: I'm agree with @dreispt 
(In fact, I don't remember the reason of use copy instead.)

JesusZapata: @dreispt @moylop260 
I refactor the method `get_server_script` and now avoid the copy
I more simple

PR #435, opened 2017-03-10
Ignore *_example packages in coverage check

simahawk:
See https://github.com/OCA/website-cms/pull/9#pullrequestreview-22552007

@moylop260 alternatively we could consider looking up for custom `.coveragerc` in the repo. What do you think?

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,7 +42,7 @@ Coveralls configuration file
 ----------------------------
 
 Coveralls provides information on the test coverage of your modules.
-Currently the Coveralls configuration is automatic (check default configuration [here]('cfg/.coveragerc')).
+Currently the Coveralls configuration is automatic (check default configuration [here](cfg/.coveragerc)).
 So, as of today, you don't need to include a `.coveragerc` into the repository,
 If you do it, it will be simply ignored.
 
--- a/README.md
+++ b/README.md
@@ -42,9 +42,13 @@ Coveralls configuration file
 ----------------------------
 
 Coveralls provides information on the test coverage of your modules.
-Currently the Coveralls configuration is automatic, so you don't need to include a `.coveragerc`
-to the repository. Please note that if you do it, it will be ignored.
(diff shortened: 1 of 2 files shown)


lasley:
I think this change will also need a corresponding note in the contribution docs?

simahawk:
@moylop260 @lasley there you go https://github.com/OCA/maintainer-tools/pull/263 ;)

dreispt:
Err, sorry, I thought that the note in the docs would go in this repo.
IMO the documentation regarding MQT should go in the MQT repo

lasley:
My thought was that _example is a convention of how to provide an example module & so we would put it in the contribution docs. Maybe both?

simahawk:
@lasley @dreispt I added a note in MQT readme and a link to MT#tests info.

pedrobaeza:
I'm not sure if this apply the same for Codecov. Anyway, we need to indicate the new coverage tool in the docs. @moylop260, you made the switch. Do you know if the file still applies?

moylop260:
FYI Codecov use the same output of coverage command using the same .coveragerc file

README.md
```diff
@@ -42,9 +42,13 @@ Coveralls configuration file
 ----------------------------
 
 Coveralls provides information on the test coverage of your modules.
-Currently the Coveralls configuration is automatic, so you don't need to include a `.coveragerc`
-to the repository. Please note that if you do it, it will be ignored.
+Currently the Coveralls configuration is automatic (check default configuration [here](cfg/.coveragerc)).
```
pedrobaeza: s/Coveralls/Codecov

simahawk: @pedrobaeza no more Coveralls? Shouldn't be "Coveralls and Codecov"?

pedrobaeza: We are extinguishing it, but we can put both in a preventive way

simahawk: updated and linked both services

PR #437, opened 2017-03-20
[FIX] Update PhantomJS

lasley:
* Add PhantomJS update to circumvent [MutationObserver](https://travis-ci.org/OCA/e-commerce/jobs/213050636#L847) error in V10

We have now used this to fix tours in 3 PRs:
* https://github.com/OCA/website/pull/329
* https://github.com/OCA/e-commerce/pull/155
* https://github.com/OCA/website/pull/331

The actual install/update adds all of 10 seconds on to the build. IMO this is better than requiring a change in relevant Travis files - this error is pretty hard to spot & took me a decent amount of time to wrestle down.

I switched OCA/website#329 to use this branch in https://github.com/OCA/website/pull/329/commits/82e5e8a06b6fb21a85e9b490d3cd44fb67473f9b to provide proof of fix.

### DIFF
--- a/travis/travis_install_nightly
+++ b/travis/travis_install_nightly
@@ -33,8 +33,7 @@ npm install -g less less-plugin-clean-css
 
 
 # Update PhantomJS (v10 compat)
-if [ "${PHANTOMJS_VERSION}" != "OS" ];
-then
+if [ "${PHANTOMJS_VERSION}" != "OS" ]; then
     npm install --prefix ${TRAVIS_BUILD_DIR} "phantomjs-prebuilt@${PHANTOMJS_VERSION:=latest}"
     ln -s "${TRAVIS_BUILD_DIR}/node_modules/phantomjs-prebuilt/lib/phantom/bin/phantomjs" "${HOME}/maintainer-quality-tools/travis/phantomjs"
 fi
--- a/travis/travis_install_nightly
+++ b/travis/travis_install_nightly
@@ -32,6 +32,11 @@ ln -s `which nodejs` $HOME/maintainer-quality-tools/travis/node
 npm install -g less less-plugin-clean-css
 
 
+# Update PhantomJS (v10 compat)
+npm install --prefix ${TRAVIS_BUILD_DIR} phantomjs-prebuilt@2.1.14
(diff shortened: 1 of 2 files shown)


lasley:
cc @laslabs @yajo

travis/travis_install_nightly
```diff
@@ -32,6 +32,11 @@ ln -s `which nodejs` $HOME/maintainer-quality-tools/travis/node
 npm install -g less less-plugin-clean-css
 
 
+# Update PhantomJS (v10 compat)
+npm install --prefix ${TRAVIS_BUILD_DIR} phantomjs-prebuilt@2.1.14
```
Yajo: We could remove the `@2.1.14` part and let it be the latest always, don't you think?
Well, I know this could lead to problems if API changes... Maybe `@2` would be enough (not sure how npm handles version strings...).

lasley: I can do `@latest`, but that's about the only wild-type version from what I understand.

lasley:
@Yajo - I think features.js is a pretty good idea. What if I just ran it immediately after the phantom upgrade? It would add a crapload to the output, but it's hidden by default anyways.

travis/travis_install_nightly
moylop260: What about use PHANTOMJS_VERSION instead?

moylop260: I mean, maybe odoo 8.0 requires a old version of phantomjs or other odoo version require a particular phantomjs version where we need use a environment variable to assign it

If the version is not assigned then use the default OS version

lasley: Using a `PHANTOMJS_VERSION` env var seems good, yeah. 

I would prefer to default upgrade though, instead of defaulting to the old version. As @Yajo mentioned, we have been held back by the old version artificially before it outright broke us.

lasley:
#438 created for features.js output

travis/travis_install_nightly
```diff
@@ -32,6 +32,13 @@ ln -s `which nodejs` $HOME/maintainer-quality-tools/travis/node
 npm install -g less less-plugin-clean-css
 
 
+# Update PhantomJS (v10 compat)
+if [ "${PHANTOMJS_VERSION}" != "OS" ];
```
Yajo: You're gonna kill me for this comment, 😆 but if you are splitting the `if` and `then` in 2 lines, you don't need the `;`

moylop260:
Fast tracking since that we have other issues related

PR #441, opened 2017-04-20
[REF] test_server: Hidden secret params

moylop260:
- Avoid show database password or `db_*` parameters


moylop260:
This fix is important in order to avoid that runbot show `--log-db` information like as:
`--log-db=postgres://{db_user}:{db_password}@{db_host}/{dbname}`

Notice `db_password` and other important information to connect directly to database.

hbrunn:
tests can come later

PR #442, opened 2017-04-20
[FIX] test_server: Support empty preinstall modules

moylop260:
Fix the following output

"""
Modules to test: base
'NoneType' object is not iterable

+=======================================
|  Tests summary:
|---------------------------------------
| test_server.py              FAIL
+=======================================
"""


bealdav:
:+1:

PR #449, opened 2017-06-16
[REF] install: Skip webkit installation if there is a issue

moylop260:
The new host server there are many network issue.
Webkit is not a package required to run the tests correctly in the major of the builds.
So we can continue if there is a issue.

### DIFF
--- a/travis/travis_install_nightly
+++ b/travis/travis_install_nightly
@@ -82,10 +82,9 @@ if [ "$clone_result" != "0"  ]; then
     exit $clone_result
 fi;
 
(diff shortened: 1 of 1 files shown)


pedrobaeza:
Can you add an environment variable to control if this is required?

dreispt:
Today all OCA TraviSCI builds are failing. This is the fix, right?
Do you know why the download fails in the first place?
I tried the URL and it seems to work for me.

moylop260:
@pedrobaeza 
Good point, thanks!
Done!

moylop260:
@dreispt yes, this is the fix

PR #450, opened 2017-06-16
[FIX] Switch to wkhtmltopdf GitHub URI

lasley:
Our CI is failing due to https://github.com/wkhtmltopdf/wkhtmltopdf/issues/3518. I propose that we switch to a Github mirror temporarily, then we can switch back once I get the notification on the ticket.

An alternative, which may actually be better, is to host our wkhtmltopdf inside of MQT & not deal with the remote downloads. This isn't the first time we've had issues with wkhtmltopdf installations.

PR #451, opened 2017-06-16
[FIX] wkhtmltox build arg

lasley:
* Add missing WKHTMLTOPDF_VERSION declaration and use a SKIP arg instead of blank

This env var was introduced in #449, but doesn't exist anywhere. 

Additionally - In order to avoid having to change all our Travis files, I propose that we skip the installation with the value `SKIP`, otherwise we install the version we have defined (`0.12.4`)


travis/travis_install_nightly
```diff
@@ -82,7 +84,7 @@ if [ "$clone_result" != "0"  ]; then
     exit $clone_result
 fi;
 
-if [ "${WKHTMLTOPDF_VERSION}" != "" ]; then
+if [ "${WKHTMLTOPDF_VERSION}" != "SKIP" ]; then
     echo "Install webkit (wkhtmltopdf) patched version ${WKHTMLTOPDF_VERSION}"
     (cd ${HOME}/maintainer-quality-tools/travis/ && wget -qO- -t 1 --timeout=240 https://downloads.wkhtmltopdf.org/0.12/${WKHTMLTOPDF_VERSION}/wkhtmltox-${WKHTMLTOPDF_VERSION}_linux-generic-amd64.tar.xz | tar -xJ --strip-components=2 wkhtmltox/bin/wkhtmltopdf)
```
moylop260: Should we change the url to use github here? (without checksum process)

lasley: I thought you said to leave that in #450? That's the only thing  that PR consists of now

lasley: The diffs should align I think regardless of merge order due to not editing the same or bordering lines?

moylop260: The travis is red IMHO we need change the url to use the new github URL (if we are forcing a installation by default)

lasley:
Alright this should be 🍏  now

lasley:
Rebased to solve conflicts.

PR #452, opened 2017-06-16
[IMP] Add checksum verification for wkhtmltox

lasley:
This adds checksum verification to the wkhtmltox download, as discussed in #450. 

Depends on:
- [ ] #450

### DIFF
--- a/travis/travis_install_nightly
+++ b/travis/travis_install_nightly
@@ -84,7 +84,7 @@ fi;
 
 if [ "${WKHTMLTOPDF_VERSION}" != "" ]; then
     echo "Install webkit (wkhtmltopdf) patched version ${WKHTMLTOPDF_VERSION}"
-    (cd ${HOME}/maintainer-quality-tools/travis/ && wget -qO- -t 1 --timeout=240 https://downloads.wkhtmltopdf.org/0.12/${WKHTMLTOPDF_VERSION}/wkhtmltox-${WKHTMLTOPDF_VERSION}_linux-generic-amd64.tar.xz | tar -xJ --strip-components=2 wkhtmltox/bin/wkhtmltopdf)
+    (cd ${HOME}/maintainer-quality-tools/travis/ && wget -qO- -t 1 --timeout=240 https://github.com/wkhtmltopdf/wkhtmltopdf/releases/download/${WKHTMLTOPDF_VERSION}/wkhtmltox-${WKHTMLTOPDF_VERSION}_linux-generic-amd64.tar.xz | tar -xJ --strip-components=2 wkhtmltox/bin/wkhtmltopdf)
 fi;
 
 # Expected directory structure:
--- a/travis/travis_install_nightly
+++ b/travis/travis_install_nightly
@@ -84,7 +84,17 @@ fi;
 
 if [ "${WKHTMLTOPDF_VERSION}" != "" ]; then
     echo "Install webkit (wkhtmltopdf) patched version ${WKHTMLTOPDF_VERSION}"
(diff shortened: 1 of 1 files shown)


travis/travis_install_nightly
```diff
@@ -84,7 +84,17 @@ fi;
 
 if [ "${WKHTMLTOPDF_VERSION}" != "" ]; then
     echo "Install webkit (wkhtmltopdf) patched version ${WKHTMLTOPDF_VERSION}"
-    (cd ${HOME}/maintainer-quality-tools/travis/ && wget -qO- -t 1 --timeout=240 https://downloads.wkhtmltopdf.org/0.12/${WKHTMLTOPDF_VERSION}/wkhtmltox-${WKHTMLTOPDF_VERSION}_linux-generic-amd64.tar.xz | tar -xJ --strip-components=2 wkhtmltox/bin/wkhtmltopdf)
+
+    cd ${HOME}/maintainer-quality-tools/travis/ &&
+    wkhtmltox=$(wget -qO- -t 1 --timeout=240 https://github.com/wkhtmltopdf/wkhtmltopdf/releases/download/${WKHTMLTOPDF_VERSION}/wkhtmltox-${WKHTMLTOPDF_VERSION}_linux-generic-amd64.tar.xz) &&
+    WKHTMLTOX_CHECKSUM='049b2cdec9a8254f0ef8ac273afaf54f7e25459a273e27189591edc7d7cf29db'
```
lasley: From  @moylop260 in  https://github.com/OCA/maintainer-quality-tools/pull/450#discussion_r122519817


> I'm not sure of use this environment variable just for webkit.
> I mean, we could use a wget_custom with a dictionary of checksum and url used from MQT.
> But IMHO this is other feature (then other PR).
> 
> Could we use just the change of the URL like as main comment?

lasley: This isn't being exported though, so it doesn't survive past the execution of this script - or am I wrong?

The alternative is just checking the sum against a string, which I do not feel is as clear.

> I mean, we could use a wget_custom with a dictionary of checksum and url used from MQT.

IMO this is another feature in itself. We currently support one version of wkhtmltopdf according to what I see here.

pedrobaeza:
I have updated the dependency to #450, as I think it's the correct one (not #50)

pedrobaeza:
Waiting @moylop260's  feedback

moylop260:
Is possible skip the checksum validation if the environment variable is missing?

lasley:
@moylop260 - Which environment variable? The checksum validation does not take place if the env var for the wkhtmltox version is `SKIP`. Is that what you're referring to?

moylop260:
Could re-use current environment of TESTS=1 from travis.yml to set the following variables?
1. WKTHMLTOPDF_VERSION = version1
2. WKTHMLTOPDF_VERSION = version1
3. and SKIP
4. w/o define

lasley:
Hey @moylop260 - sorry, but I'm not sure I understand. Are you proposing that we just default to the `WKTHMLTOPDF_VERSION` when `TEST=1`, but otherwise default to skip - if the variable isn't otherwise declared? So something like this?

| `TESTS` | `WKTHMLTOPDF_VERSION` defined | `WKTHMLTOPDF_VERSION` actual |
|--------|------------------------------------|----------------------------------|
| 1 | | $CURRENT_VERSION |
| 1 | 1.2.3 | 1.2.3 |
| 0 | | SKIP |
| 0 | 1.2.3 | 1.2.3 |

moylop260:
sorry for confusion
I meant that we should increase the coverage for this script using all conditions.
So we should change the .travis.yml to test all cases 

I'm not talking about a change of behaviour

lasley:
@moylop260 - Hmmm so you're proposing that we test an invalid checksum? Do we have a way to check if something is installed in our environment using our Travis file?

lasley:
@moylop260 - I am unclear on how to proceed on this. Is there a problem with my current implementation?

moylop260:
Is there a way to skip checksum validation if is not defined?

lasley:
We could do that by just not checking the checksum if `WKHTMLTOX_CHECKSUM`, but I think that's opening up a security issue that is being closed here. What's the motivation behind no checksum while still installing wkhtmltox?

moylop260:
We will need change all `.travis.yml` files that don't have this checksum value filled yet.
Then if we merge it this PR all builds will be fail.

PR #453, opened 2017-06-21
Add helper comment to install a specific wkhtmltopdf version

jbeficent:
Sometimes it is important to install a specific wkhtmltopdf on a repo. Maintainers need to know how to properly set this up.

See https://github.com/OCA/reporting-engine/pull/144#discussion-diff-123189206R32 for an example.


jbeficent:
cc @etobella

PR #454, opened 2017-07-03
[IMP] apis.py: Adding new method _component_commit

JesusZapata:
This method make commit to the uncommitted changes on weblate repository

In some cases the user does not make the commit into the repository of weblate.
Because the user do not have permission or because the user does not know that it is a necessary action in the process of the translated
With this change the commit will do if necessary

Related with https://github.com/Vauxoo/maintainer-quality-tools/pull/213

### DIFF
--- a/travis/apis.py
+++ b/travis/apis.py
@@ -129,11 +129,10 @@ class WeblateApi(Request):
         needs_commit = self._request(url)
         if not needs_commit['needs_commit']:
             return
-        for i in range(10):
-            new_commit = self._request(url, {'operation': 'commit'})
-            if new_commit['result']:
-                break
-            time.sleep(60)
+        new_commit = self._request(url, {'operation': 'commit'})
+        if not new_commit['result']:
+            raise ApiException('The commit into the component "%s" cannot be '
+                               'made' % component['slug'])
         return True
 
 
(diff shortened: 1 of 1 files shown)


JesusZapata:
@moylop260 
In this PR MQT check if the repository needs commit and if need then made one commit using the API

travis/apis.py
```diff
@@ -122,6 +123,19 @@ def _component_lock(self, component, lock=True):
             time.sleep(60)
         return True
 
+    def _component_commit(self, component):
+        url = (self.host + '/components/%s/%s/repository/' %
+               (self.project['slug'], component['slug']))
+        needs_commit = self._request(url)
+        if not needs_commit['needs_commit']:
+            return
+        for i in range(10):
```
pedrobaeza: Why this repeat loop?

JesusZapata: The loop is only one validation extra, if the api can not make the commit or another trouble related with the API
If all goes well the loop only make one iteration

pedrobaeza: Yeah, but iterating several times is not going to be the solution. If the commit can't be done, the result will be the same after 10 times. I know that you think in connection problems, but check this, not a general try.

JesusZapata: I deleted the loop and add a `ApiException` if the commit cannot be made

PR #456, opened 2017-07-06
[FIX] travis_weblate.py: Adding more patch to de odoo

JesusZapata:
I ported the changes made in https://github.com/odoo/odoo/pull/17925 over the files `odoo/addons/base/ir/ir_translation.py` and `odoo/tools/translate.py`

Now MQT apply these patch if the patch has not been applied

The change in the file `(odoo|openerp)/addons/base/ir/ir_translation.py` 

image

The change in the file `(odoo|openerp)/tools/translate.py` 

image

### DIFF
--- a/travis/travis_weblate.py
+++ b/travis/travis_weblate.py
@@ -5,7 +5,6 @@ import os
 import re
 import glob
 import subprocess
-import shlex
 
 from odoo_connection import Odoo10Context, context_mapping
 from test_server import (get_addons_path, get_server_path, parse_list)
@@ -78,19 +77,10 @@ class TravisWeblateUpdate(object):
                 subprocess.call(sed)
             p_file = os.path.join(self._server_path, base,
                                   os.path.join('addons', 'base', 'ir',
-                                               'ir_translation_test.py'))
+                                               'ir_translation.py'))
             if os.path.isfile(p_file):
-                rgrep = ["rgrep", "-h", "-n", "'if not trans_dict'",
-                         p_file]
-                print " ".join(rgrep)
-                p = subprocess.Popen(shlex.split(" ".join(rgrep)),
-                                     stdout=subprocess.PIPE)
-                out = p.communicate()
-                if not (out and out[0] and out[0].split(':')):
-                    return
-                line = int(out[0].split(':')[0])
-                sed = ["sed", "-i", "-e", r"%s,%sd" % (line, line + 1),
-                       p_file]
+                sed = ["sed", "-i", "-e",
+                       r"s/if\snot\strans_dict\['value']:/if False:/g", p_file]
                 print " ".join(sed)
                 subprocess.call(sed)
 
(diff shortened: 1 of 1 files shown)


JesusZapata:
@moylop260 
I applied the changes made on https://github.com/odoo/odoo/pull/17925 over MQT

travis/travis_weblate.py
```diff
@@ -67,17 +68,31 @@ def _apply_patch_odoo(self):
...
-                   r"s/translation'] = src/translation'] = ''/g",
-                   s_file]
-            print " ".join(cmd)
-            subprocess.call(cmd)
+        for base in ('odoo', 'openerp'):
+            p_file = os.path.join(self._server_path, base,
+                                  os.path.join('tools', 'translate.py'))
+            if os.path.isfile(p_file):
+                sed = ["sed", "-i", "-e",
+                       r"s/translation'] = src/translation'] = ''/g", p_file]
```
moylop260: You can avoid the loop `for base in ('odoo', 'openerp')` using directly `base = "{odoo,openerp}"` to cover both folders.

travis/travis_weblate.py
```diff
@@ -67,17 +68,31 @@ def _apply_patch_odoo(self):
...
+                print " ".join(sed)
+                subprocess.call(sed)
+            p_file = os.path.join(self._server_path, base,
+                                  os.path.join('addons', 'base', 'ir',
+                                               'ir_translation_test.py'))
+            if os.path.isfile(p_file):
+                rgrep = ["rgrep", "-h", "-n", "'if not trans_dict'",
+                         p_file]
+                print " ".join(rgrep)
+                p = subprocess.Popen(shlex.split(" ".join(rgrep)),
```
moylop260: Use directly rgrep without split join

travis/travis_weblate.py
```diff
@@ -67,17 +68,31 @@ def _apply_patch_odoo(self):
...
+            if os.path.isfile(p_file):
+                sed = ["sed", "-i", "-e",
+                       r"s/translation'] = src/translation'] = ''/g", p_file]
+                print " ".join(sed)
+                subprocess.call(sed)
+            p_file = os.path.join(self._server_path, base,
+                                  os.path.join('addons', 'base', 'ir',
+                                               'ir_translation_test.py'))
+            if os.path.isfile(p_file):
+                rgrep = ["rgrep", "-h", "-n", "'if not trans_dict'",
```
moylop260: You can use `sed -i '/pattern/d'` to delete a line and avoid a rgrep

travis/travis_weblate.py
JesusZapata: @moylop260 
Thanks for teach me this way to use multiple folders in `rgrep` command
I try to use this example into the python and i can not made to work

image

I the correct way to run this command ?

moylop260:
We need create a full POC with this change...
If there is a change to create we will use a new PR

PR #457, opened 2017-07-06
[REF] install: Install requests[security] again to avoid new errors

moylop260:


pedrobaeza:
I'm going to fast-track this for checking if l10n-spain is fixed.

PR #209, opened 2015-07-05
[IMP] pip --user argument to make travis .NET build work

mathi123:
this fixes my install bugs for odoo + .NET build with mono, on 1 system
e.g. see [here](https://travis-ci.org/mathi123/dotnet/builds/68430398)

PR #458, opened 2017-07-06
[REF] install: pip install w/o --user

moylop260:
Travis use a virtualenv to install python packages, if we use `pip install --user ...` then it will have 2 environments with possible errors like as:
https://github.com/OCA/l10n-spain/pull/569

Before this PR:
 - screen shot 2017-07-06 at 17 02 35

After this PR:
 - screen shot 2017-07-06 at 17 25 08

The `--user` parameters was added https://github.com/OCA/maintainer-quality-tools/pull/209 because a error from https://github.com/OCA/dotnet/pull/6 but that environment is not using a standard python image.

They are using:
https://docs.travis-ci.com/user/languages/csharp/
And the log show us: https://travis-ci.org/OCA/dotnet/builds/70875564#L113

```bash
C# support for Travis-CI is community maintained.
Please open any issues at https://github.com/travis-ci/travis-ci/issues/new and cc @joshua-anderson @akoeplinger @nterry
```

Then is not a standard build support by travis and this shouldn't a standard build support by us, in fact, that project is not compatible with OCA/MQT since too much previous commits.

Fix https://github.com/OCA/maintainer-quality-tools/issues/459

### DIFF
--- a/travis/travis_install_nightly
+++ b/travis/travis_install_nightly
@@ -68,7 +68,7 @@ fi
 # Workaround to force using system site packages (see https://github.com/Shippable/support/issues/241#issuecomment-57947925)
 rm -f $VIRTUAL_ENV/lib/python2.7/no-global-site-packages.txt
 pip install --upgrade pip
-pip install --user -q --no-binary pycparser -r ${HOME}/maintainer-quality-tools/requirements.txt
+pip install -q --no-binary pycparser -r ${HOME}/maintainer-quality-tools/requirements.txt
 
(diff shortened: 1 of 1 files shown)


pedrobaeza:
Can I fast-track?

lasley:
Is this the solution to #459? The error seems totally relevant to this

moylop260:
> Is this the solution to #459? The error seems totally relevant to this

Yes is it

moylop260:
> Can I fast-track?

Yes please

PR #465, opened 2017-07-19
[IMP] Add shortcut to install SASS and Compass

lasley:
This adds a shortcut for the installation of website dependencies SASS and Compass, which has been causing some considerable issues lately. Most recent 
* https://github.com/OCA/website/pull/361#issuecomment-316003296
* https://github.com/OCA/social/pull/179#issuecomment-316238557
cc @moylop260 @Yajo

### DIFF
--- a/travis/travis_install_nightly
+++ b/travis/travis_install_nightly
@@ -32,10 +32,9 @@ ln -s `which nodejs` $HOME/maintainer-quality-tools/travis/node
 npm install -g less less-plugin-clean-css
 
 if [ "${WEBSITE_REPO}" == "1" ]; then
-    rvm install ruby 2.0 &&
-    rvm use 2.0 &&
-    gem install bootstrap-sass &&
-    gem install compass --pre
+    rvm install ruby --latest
+    rvm use ruby --latest
+    gem install compass bootstrap-sass
 fi
 
 # Update PhantomJS (v10 compat)
--- a/travis/travis_install_nightly
+++ b/travis/travis_install_nightly
@@ -31,6 +31,12 @@ fi
 ln -s `which nodejs` $HOME/maintainer-quality-tools/travis/node
 npm install -g less less-plugin-clean-css
 
+if [ "${WEBSITE_REPO}" == "1" ]; then
+    rvm install ruby 2.0 &&
+    rvm use 2.0 &&
+    gem install bootstrap-sass &&
+    gem install compass --pre
+fi
 
 # Update PhantomJS (v10 compat)
 if [ "${PHANTOMJS_VERSION}" != "OS" ]; then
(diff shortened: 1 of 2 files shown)


travis/travis_install_nightly
```diff
@@ -31,6 +31,12 @@ fi
 ln -s `which nodejs` $HOME/maintainer-quality-tools/travis/node
 npm install -g less less-plugin-clean-css
 
+if [ "${WEBSITE_REPO}" == "1" ]; then
+    rvm install ruby 2.0 &&
+    rvm use 2.0 &&
+    gem install bootstrap-sass &&
```
Yajo: This file starts with `set -e` which will cause it to be aborted on any error, so you don't need those `&&`

Yajo:
It seems in my tests that if we `gem install bootstrap-sass compass sass`, we end up with 2 versions of `sass`: 3.4.25 and 3.5.1 (bleeding edge). If you uninstall 3.5.1, it works fine.

So, we should just:

```
rvm install ruby --latest
rvm use ruby --latest
gem install bootstrap-sass compass
```

If I'm not wrong, that should do the trick.

Yajo:
I was not so far from the solution. The problem is:

- [bootstrap-sass requires sass >= 3.3.4](https://github.com/twbs/bootstrap-sass/blob/master/bootstrap-sass.gemspec#L14)
- [compass requires sass >= 3.3.0 and < 3.5](https://github.com/Compass/compass/blob/stable/core/compass-core.gemspec#L24)

So if you install before bootstrap-sass, it will download the latest version (3.5.1), and when you install compass later, it will need to download the latest 3.4.* version (3.4.25). There you yave: 2 sass versions, one with bootstrap-sass and the other with compass enabled, but none with both.

Besides, there's the known problem of having to use ruby>=2.0, so the final fix is to install compass before bootstrap-sass:

```
rvm install ruby --latest
rvm use ruby --latest
gem install compass bootstrap-sass # Notice compass goes before!
```

lasley:
Updated, thanks @Yajo

Yajo:
Please somebody merge this! 😭

treviser:
FYI: I have made a little HowTo on how to install the proper SASS components for Odoo on Ubuntu 16.04 using the Brightbox repository here: https://www.odoo.com/forum/help-1/question/ubuntu-16-04-how-to-install-sass-for-odoo-123090#answer-123095
I did find that the method using rvm was very difficult (because I did not succeed in properly installing rvm in the first place) and this works perfectly for me.

lasley:
Thanks @treviser - this was indeed a bitch.

@Yajo - are we having any more issues, or is it finally solved? I haven't seen anything recently & I hesitate to touch this house of cards without reason.

Yajo:
It seems to be working recently

PR #469, opened 2017-07-21
[FIX] travis_weblate.py: Create folder i18n and *.po file

JesusZapata:
When one module is not into the list of component on weblate, that module is excluded for `travis_weblate.py`

In this pull request this mistage was fixes up.
Based on the modules installed and the components registered on weblate, if that module is installed but the same module is not present into the weblate component now the folder `i18n` and the *.po files is going to created

Only the environment variable `LANG_ALLOWED` should be setter to a default list of languages supported

Related with https://github.com/Vauxoo/maintainer-quality-tools/pull/219


JesusZapata:
@moylop260 
In this PR has the same solution of https://github.com/Vauxoo/maintainer-quality-tools/pull/219

PR #475, opened 2017-08-04
[FIX] Only source rvm script if avail

lasley:
* Source RVM from conditional paths (support T2D & Travis)

Testing in https://github.com/OCA/e-commerce/pull/194

### DIFF
--- a/travis/travis_install_nightly
+++ b/travis/travis_install_nightly
@@ -31,7 +31,11 @@ ln -s `which nodejs` $HOME/maintainer-quality-tools/travis/node
 npm install -g less less-plugin-clean-css
 
 if [ "${WEBSITE_REPO}" == "1" ]; then
-    source ~/.rvm/scripts/rvm
+    if [ -f ~/.rvm/scripts/rvm ]; then
+        source ~/.rvm/scripts/rvm
+    else
+        source /usr/local/rvm/scripts/rvm
+    fi
     rvm install ruby --latest
     rvm use ruby --latest
     gem install compass bootstrap-sass


lasley:
Alright this works

pedrobaeza:
Is this the definitive solution? Can I merge?

lasley:
Let's wait for the two builds in https://github.com/OCA/e-commerce/pull/194 to finish on OCA side, but Travis and Runbot both passed on my side.

lasley:
Note the issue was a difference of where the RVM script is stored in the two environments, so fairly simple once @moylop260 showed me which image we're using. 

I think @Yajo may have a point about us needing to move to a fully Dockerized Travis build, instead of our current strategy. I feel like we're going to keep finding more and more things like this each time Travis shifts their build image from under us

pedrobaeza:
Yeah, I think we can move our docker-odoo-base project to OCA and use it for everything as it already contains all the needed things for making Odoo to work. We are already starting a PoC for using it in our Gitlab for CI (Gitlab runner).

lasley:
It'd be nice to see the base image in OCA. It's working well for us in numerous production environments at this point, so it makes sense to have it under the umbrella.

lasley:
Also we're good for a merge on this - Runbot and Travis are 🍏  on our side

moylop260:
> I think @Yajo may have a point about us needing to move to a fully Dockerized Travis build, instead of our current strategy. I feel like we're going to keep finding more and more things like this each time Travis shifts their build image from under us

I'm agree with that.

> Yeah, I think we can move our docker-odoo-base project to OCA and use it for everything as it already contains all the needed things for making Odoo to work. We are already starting a PoC for using it in our Gitlab for CI (Gitlab runner).

In fact, 2 years ago I tried create it from https://github.com/OCA/maintainer-quality-tools/pull/152 but that was not welcome.

The following task are:
 - Create a oca/docker-odoo-base project and a very vey basic Dockerfile with minimal requirements to work with odoo like as https://github.com/OCA/maintainer-quality-tools/pull/152  but updated.
 - Create a MQT script to use docker (I'm not sure if we can use t2d, but could be a easiest implementation to start)
 - Create dummy PR for main projects to test it

moylop260:
Maybe is a work for our next OCA's meeting in Belgium

pedrobaeza:
Yeah, I mean to change .travis.yml by a very small one (the one that loads docker-odoo-base Docker), and then use the powerful scaffolding system we have in it to accomplish any pip/apt/npm/... dependency, any other repository dependencies, etc, with it.

It's a big change, but we will remove all these problems at once. And even better, we will have the same environment for automated tests, runbot builds, development and production!!

lasley:
> And even better, we will have the same environment for automated tests, runbot builds, development and production!!

Don't forget [Runbot itself](https://blog.laslabs.com/2017/07/deploying-fully-dockerized-runbot/)!

Issue #447, opened 2017-06-01
Incorrect Build Success


dreispt:
This build should have failed, but passed:
https://travis-ci.org/OCA/project/jobs/237996931#L1107

dreispt:
I guess thus @dreispt guy is to [blame](https://github.com/OCA/maintainer-quality-tools/commit/ee9aef3650b3c08922bc363c96612ac3dd2b8d70).
For reference, #43 motivated the change.

dreispt:
@sbidoul [This](https://travis-ci.org/OCA/management-system/jobs/215216204#L550) build has the same missing dependency error, yet it fails!

It is two months old. Maybes something changed in between? 🤔 
Not the code you pointed to, surely...

dreispt:
Might be related to https://github.com/OCA/maintainer-quality-tools/commit/be28ef3ec0cb13c96b50b0a639f1bc316c64120c
@pedrobaeza should it be reverted?

sbidoul:
@dreispt in my experience the Odoo exit code is not reliable. So detecting ERROR would be a good thing to do IMO.

moylop260:
What about use the [runbot's regular expression](https://github.com/odoo/odoo-extra/blob/9e1ff15449ba56fdf1d3cc8993dec428da54c8dd/runbot/runbot.py#L46-L47) but for MQT?

```python
_re_error = r'^(?:\d{4}-\d\d-\d\d \d\d:\d\d:\d\d,\d{3} \d+ (?:ERROR|CRITICAL) )|(?:Traceback \(most recent call last\):)$'
_re_warning = r'^\d{4}-\d\d-\d\d \d\d:\d\d:\d\d,\d{3} \d+ WARNING '
```

We will have all cases covered.

moylop260:
Using the same output from https://travis-ci.org/OCA/project/jobs/237996931#L1107

The result is:

`python -c "import re;_re_error = r'^(?:\d{4}-\d\d-\d\d \d\d:\d\d:\d\d,\d{3} \d+ (?:ERROR|CRITICAL) )|(?:Traceback \(most recent call last\):)$';regexp = re.compile(_re_error, re.M);print '\n'.join(regexp.findall(open('job_test_all.txt').read()))"`

```bash
2017-05-31 23:36:23,512 5776 ERROR
2017-05-31 23:36:38,510 5776 ERROR
2017-05-31 23:36:38,514 5776 ERROR
```

For all odoo's versions

[...]

PR #476, opened 2017-08-08
[FIX] test_server.py: Adding more validation to detected ERROR and WARNING

JesusZapata:
Relating to https://github.com/OCA/maintainer-quality-tools/issues/447#issuecomment-311502587

With this small change now travis is red if in the log has one `ERROR`

I made one PR of test related with this https://github.com/JesusZapata/project/pull/1

image

And now the travis is red

### DIFF
--- a/travis/test_server.py
+++ b/travis/test_server.py
@@ -32,6 +32,7 @@ def has_test_errors(fname, dbname, odoo_version, check_loaded=True):
         ]
     errors_report = [
         lambda x: x['loglevel'] == 'CRITICAL',
+        lambda x: x['loglevel'] == 'ERROR',
         'At least one test failed',
         'no access rules, consider adding one',
         'invalid module names, ignored',


JesusZapata:
@dreispt @moylop260 @sbidoul 
Yours comments are welcome

dreispt:
Sometimes we test if an error is raised.
Will this incorrectly break such a build?

JesusZapata:
@dreispt 
The execution of odoo continues running without paying attention to the error
I don't know if this behaviour is the correct

sbidoul:
I also noticed that, but I never dared to change it, as I supposed that if someone explicitly filtered on CRITICAL errors only, there must have been a reason. 

FWIW, for our own CI needs, we created a script [dedicated to checking odoo logs](https://github.com/acsone/acsoo/blob/master/acsoo/checklog.py), with agressive defaults (WARNING level records are considered errors) and a regex-based mechanism to filter expected errors.

dreispt:
To be clear, I'm not against this change as long as long as tests verifying if a Exception was raised don't incorrectly break the build.
Can you check that @JesusZapata ?

nhomar:
> FWIW, for our own CI needs, we created a script dedicated to checking odoo logs, with agressive defaults (WARNING level records are considered errors) and a regex-based mechanism to filter expected errors.

@sbidoul that's more or less what we do internally hacking a little mqt itself.

@moylop260 is this PR valid to be merged as it is? I am worried of change the level and bring false green/red on this.

I prefer move this to a future version, What do you think?

lasley:
Hi @JesusZapata & @moylop260 - any update on this?

PR #478, opened 2017-08-11
[FIX] install: Uninstall current gem packages

moylop260:
In order to fix issues related with double versions of gem packages
Fix https://github.com/OCA/maintainer-quality-tools/issues/471


moylop260:
IMHO this could be merged
More info about https://github.com/OCA/maintainer-quality-tools/issues/471#issuecomment-321888053

PR #479, opened 2017-08-30
[IMP] clean up wkhtmltopdf configuration

hbrunn:
removed some superfluous config settings regarding wkhtmltopdf, by now we install this in `travis_install_nightly`, no apt necessary and on newer build environments (trusty), installing via apt is actually harmful because it's the wrong version but wins against the version installed by said script.

Used in https://github.com/OCA/reporting-engine/pull/141

### DIFF
--- a/sample_files/.travis.yml
+++ b/sample_files/.travis.yml
@@ -10,9 +10,7 @@ python:
 
 addons:
   apt:
-# only add the two lines below if you need wkhtmltopdf for your tests
 #    sources:
-#      - pov-wkhtmltopdf
 #    Search your sources alias here:
 #      https://github.com/travis-ci/apt-source-whitelist/blob/master/ubuntu.json
     packages:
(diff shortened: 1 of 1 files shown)


sample_files/.travis.yml
```diff
@@ -25,11 +23,6 @@ addons:
 #       https://github.com/travis-ci/apt-package-whitelist/blob/master/ubuntu-precise
 #      - wkhtmltopdf  # only add if needed and check the before_install section below
 
-# set up an X server to run wkhtmltopdf.
```
pedrobaeza: Is this not needed anymore?

hbrunn: fortunately not with the patched version. Check the PR above, now the build is finally green with this :tada:

pedrobaeza:
Should we add `dist: trusty` also here?

hbrunn:
according to https://docs.travis-ci.com/user/reference/overview/, this will happen automatically eventually. I forced it now for reporting-engine because I thought my travis problems were due to the ancient ubuntu, and now I keep it there in order not to get any bad surprises. But adding it here might cause bad surprises in other repos so I think better not.

pedrobaeza:
OK, thanks for the explanations.

PR #481, opened 2017-08-30
[IMP] maintainer-quality-tools: Enable installing 0.12.1 version of webkit qt patched

mapuerta:
This PR add support for version >= 0.12.1 of webkit qt patched

This change was suggested in this issue https://github.com/Vauxoo/maintainer-quality-tools/issues/227

### DIFF
--- a/travis/travis_install_nightly
+++ b/travis/travis_install_nightly
@@ -124,11 +124,24 @@ if [ "$clone_result" != "0"  ]; then
     exit $clone_result
 fi;
 
-if [ "${WKHTMLTOPDF_VERSION}" != "" ]; then
-    echo "Install webkit (wkhtmltopdf) patched version ${WKHTMLTOPDF_VERSION}"
-    (cd ${HOME}/maintainer-quality-tools/travis/ && wget -qO- -t 1 --timeout=240 https://github.com/wkhtmltopdf/wkhtmltopdf/releases/download/${WKHTMLTOPDF_VERSION}/wkhtmltox-${WKHTMLTOPDF_VERSION}_linux-generic-amd64.tar.xz | tar -xJ --strip-components=2 wkhtmltox/bin/wkhtmltopdf)
+if [ "${WKHTMLTOPDF_VERSION}" == "" ]; then
+    WKHTMLTOPDF_VERSION="0.12.3"
 fi;
+echo "Install webkit (wkhtmltopdf) patched version ${WKHTMLTOPDF_VERSION}"
 
+if [[ ${WKHTMLTOPDF_VERSION} == "0.12.1" ]]
+then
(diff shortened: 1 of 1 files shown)


moylop260:
@mapuerta could you sign the CLA, please?

mapuerta:
@moylop260 Yes, I already signed CLA

moylop260:
Could you check the travis red result, please?

moylop260:
@pedrobaeza Could you review if the CLA was signed, please?

@lasley @Yajo what do you think about?

Yajo:
I think that we should support only 1 wkhtmltopdf version.

lasley:
> I think that we should support only 1 wkhtmltopdf version.

Do Odoo 7 & 8 work with the new version? I know 9 and 10 does.

@mapuerta - what is your use case for the old version? Contrary to the Odoo docs, v0.12.4 is working better for me than any of the other versions have.

moylop260:
@Yajo 
Currently we are supporting many wkhtmltopdf versions from environment variable. (One by default)

There are small differences with a report pdf format printed between wkhtmltopdf official version documented from odoo and other one.

wkhtmltopdf Official documented version v0.12.1:
 - factura

wkhtmltopdf Official documented version v0.12.3:
 - screen shot 2017-09-05 at 12 05 47

Of course, we could change the source to adjust it for 0.12.3 version but now using 0.12.1 version is wrong with small changes.

There are many python packages with newest versions (and better) but we are using the official ones from [odoo requirements.txt](https://github.com/odoo/odoo/blob/10.0/requirements.txt)

Shouldn't choose the same version from odoo documentation show 0.12.1?
 - https://github.com/odoo/odoo/blob/10.0/doc/setup/install.rst#packaged-installers

lasley:
> Of course, we could change the source to adjust it for 0.12.3 version but now using 0.12.1 version is wrong with small changes.

Why 0.12.3? Are you experiencing the same issue on the current version, 0.12.4? This version is not included in the Odoo docs, so I don't think they've tested it, but I've moved all my customers to it without any issue reports.

> Shouldn't choose the same version from odoo documentation show 0.12.1?

They're not even pointing to 0.12.1.2 in their docs, which resolves some [bad security issues](https://www.openssl.org/news/secadv/20141015.txt). IMO in this case their docs are incorrect.

Yajo:
We use currently 0.12.4 because we have experienced many problems with any other version. Besides, it makes not much sense to support one version in some reports and other version in others, since Odoo can only use 1 version at a time.

I'd go for 0.12.4 and let incompatible reports expose their incompatibility. Otherwise they'll never get fixed.

This is a test environment after all, you can use the version you like in your deployments.

PR #482, opened 2017-09-01
[FIX] apis.py: Fix when the components on weblate are more 25

JesusZapata:
When the components it is more that 25 the API of Weblate paginated the results, then should be get each page

For more information https://docs.weblate.org/en/latest/api.html#any--

Related with https://github.com/Vauxoo/maintainer-quality-tools/pull/231

### DIFF
--- a/travis/apis.py
+++ b/travis/apis.py
@@ -78,9 +78,17 @@ class WeblateApi(Request):
 
     def get_components(self):
         components = []
-        values = self._request(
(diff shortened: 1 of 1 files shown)


JesusZapata:
@moylop260 
In this PR has the same solution of another one PR https://github.com/Vauxoo/maintainer-quality-tools/pull/231

moylop260:
Merging since that we are testing weblate

PR #484, opened 2017-09-18
[FIX] apis.py: Adding new metod _ssh_keyscan to run the ssh-keyscan over the weblate-ssh server

JesusZapata:
When is it used the command `git clone` over the ssh the command need a console to answer the follow question

`Are you sure you want to continue connecting (yes/no)? `

With the command `ssh-keyscan` get the key a Weblate ssh server and skip the previous question


JesusZapata:
@moylop260 
This PR is necessary to put in production the Weblate + Runbot integration

PR #485, opened 2017-09-19
[IMP] Alow excluding specified modules from test_flake8 using env var

tomeyro:
With this MR we will be able to use the environment variable `EXCLUDE_FLAKE8` to specify modules to be excluded from the flake8 checks.

Use case: sometimes we need to ignore the warnings for some modules because they are from a third party and we don't/can't maintain them.

Example:

image

Here we are getting a lot of warnings from some third party modules, but we can skip them like this:

image

And then we see the warnings for the modules that correspond to us to maintain, and our CI tests don't display the false alarms.

### DIFF
--- a/travis/test_flake8
+++ b/travis/test_flake8
@@ -11,6 +11,9 @@ flake8_config_dir = os.path.join(root_dir, 'cfg')
 folders = (os.environ.get("INCLUDE_LINT", "").split() or
            get_modules(os.path.abspath('.')))
 
+exclude = os.environ.get('EXCLUDE', '').split(',')
+folders = [folder for folder in folders if folder not in exclude]
+
 status = 0
 
 for addon in folders:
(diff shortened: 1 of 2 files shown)


travis/test_flake8
```diff
@@ -11,7 +11,11 @@ flake8_config_dir = os.path.join(root_dir, 'cfg')
 
 status = 0
 
-for addon in get_modules(os.path.abspath('.')):
```
moylop260: Use same env var that pylint

tomeyro: Done.

tomeyro:
@moylop260 Now the env var is `EXCLUDE_LINT`.

image

pedrobaeza:
Why not using same variable `EXCLUDE` than TEST phase?

lasley:
Hi @tomeyro - any update on this? Are the suggestions clear?

tomeyro:
@lasley Yes. Sorry for the delay.

@moylop260 Now the env var name is `EXCLUDE` and it works for both `flake8` and `pylint` tests.

- pylint

tomeyro:
@lasley Yes. Sorry for the delay.

@moylop260 Now the env var name is `EXCLUDE` and it works for both `flake8` and `pylint` tests.

- `pylint`

image
image

- `flake8`

image
image

moylop260:
@pedrobaeza 
Could you check it again, please?

PR #486, opened 2017-09-19
[FIX] apis.py: Fix the regular expression to detect the domain containing '-'

JesusZapata:
Change the regular expression to support domain containing '-' equal to `translation.odoo-community.org`

image


JesusZapata:
@moylop260 
What do you think about this change?

PR #490, opened 2017-09-29
[REF] MQT: Migration to python3

moylop260:
[REF] MQT: Migration to python3
  - [REF] remove relative import and reduce expected lint errors
  - [REF] py3: xmlrpclib compatilibity
  - [FIX] test_server: Write bytes for binary file and print string
  - [REF] Use br for regex

[REF] MQT: Compatibility with new api odoo

  - [REF] compatible with old and new api
  - [REF] remove deprecated _columns
  - [REF] .travis.yml: Remove 6.1 and 7.0 odoo test since that 11 version don't support _columns and 6.1 and 7.0 just suppport _columns

[REF] .travis.yml: Enable py3 environment

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -31,31 +31,31 @@ matrix:
     - python: 2.7
       env:
         VERSION="9.0" TESTS="1" LINT_CHECK="0"
-        TESTS="1" INCLUDE="test_module,second_module" ODOO_REPO="OCA/OCB" MQT_TEMPLATE_DB='mqt_odoo_template_core' MQT_TEST_DB='mqt_odoo_core'
+        INCLUDE="test_module,second_module" ODOO_REPO="OCA/OCB" MQT_TEMPLATE_DB='mqt_odoo_template_core' MQT_TEST_DB='mqt_odoo_core'
     - python: 2.7
       env:
-        VERSION="9.0" TESTS="1" LINT_CHECK="0"
-        LINT_CHECK="1" TESTS="0" PYLINT_EXPECTED_ERRORS="19" TRAVIS_PULL_REQUEST="false" # Use main pylint config file
+        VERSION="9.0" TESTS="0" LINT_CHECK="1"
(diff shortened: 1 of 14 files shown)


.travis.yml
```diff
@@ -8,35 +8,54 @@ cache:
...
+      env:
+        VERSION="9.0" TESTS="1" LINT_CHECK="0"
+        VERSION="" LINT_CHECK="1" TESTS="0" PYLINT_EXPECTED_ERRORS="19" TRAVIS_PULL_REQUEST="false"  # To check VERSION empty or unset case.
+    - python: 2.7
+      env:
+        VERSION="9.0" TESTS="1" LINT_CHECK="0"
+        INCLUDE="broken_no_access_rule" SERVER_EXPECTED_ERRORS="1"  # test errors are detected
+    - python: 3.4
+      env:
+        VERSION="9.0" TESTS="1" LINT_CHECK="0"
```
pedrobaeza: Version 9.0?

moylop260: I'm not changing that part:
 - https://github.com/OCA/maintainer-quality-tools/blob/164f6801de47dfff42045a6c4216924621e97e1b/.travis.yml#L26



I just changed to use multi-python environment.

pedrobaeza: Yeah, but 9.0 doesn't work with Python 3

moylop260: Oh, I got it your point.
Thanks for clarify.

We had a global configuration using `VERSION=9.0` but I didn't find a way to use a `global` configuration using multi-python platform. (overwritten from particular one)
Then I just copied and paste in one line the global configuration in order to see the same global effect and overwrite in the next line the particular one.

What do you think about?

moylop260: @pedrobaeza 
I just removed "global" environment variable and I'm using the required.

lepistone:
@moylop260 are you sure travis doesn't support Python 3.5? From https://docs.travis-ci.com/user/languages/python/#Specifying-Python-versions it looks like they do. I'd change that to 3.5 which is the version supported by Odoo. Thanks!

pedrobaeza:
Merging as this has been green on VX branch.

PR #491, opened 2017-10-02
[REF] travis_install: Overwrite wkhtmltopdf if is installed

moylop260:
This PR is related with many false red because many projects don't have assigned the environment variable to install wkhtmltopdf patched version but the .travis.yml has `apt-get install wkhtmltopdf` so is failing because is not patched.

We can change all OCA projects to assign the environment varibable correctly but we can avoid many false red detecting if currently is installed then overwrite it.

What do you think about?

cc @Yajo @lasley @gurneyalex @sylvain-garancher

### DIFF
--- a/travis/travis_install_nightly
+++ b/travis/travis_install_nightly
@@ -124,6 +124,11 @@ if [ "$clone_result" != "0"  ]; then
     exit $clone_result
 fi;
 
+if [ "${WKHTMLTOPDF_VERSION}" == "" && $(which wkhtmltopdf) != ""]; then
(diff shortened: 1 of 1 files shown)


moylop260:
@Yajo @lasley @gurneyalex @sylvain-garancher 
This issue is related with many false red because many projects don't have assigned the environment variable to install wkhtmltopdf patched version but the .travis.yml has `apt-get install wkhtmltopdf` so is failing because is not patched.

We can change all OCA projects to assign the environment varibable correctly but we can avoid many false red detecting if currently is installed then overwrite it.

What do you think about?

gurneyalex:
I'm fine with this, as it will save tons of time. Thanks

PR #492, opened 2017-10-03
Make condition more robust

guewen:
When wkhtmltopdf is not installed, the which command is empty therefore
the comparison fails with:

line 127: [: missing `]'

Also use '-a' instead of && which is recommended by SpellCheck (https://www.shellcheck.net/2

Example of failure in the wild: https://travis-ci.org/OCA/connector/jobs/282791133#L700

### DIFF
--- a/travis/travis_install_nightly
+++ b/travis/travis_install_nightly
@@ -124,7 +124,7 @@ if [ "$clone_result" != "0"  ]; then
     exit $clone_result
 fi;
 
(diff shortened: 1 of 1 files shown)


guewen:
@moylop260 I see you committed https://github.com/OCA/maintainer-quality-tools/commit/218901354f49e169e231125b29b870ddc28f9e33 meanwhile, I don't think the double bracket is the issue (and AFAIK only supported by bash).

moylop260:
Wait... I'm not sure...

screen shot 2017-10-03 at 18 43 18
screen shot 2017-10-03 at 18 42 47

PR #493, opened 2017-10-05
fix unbuffer issue

gurneyalex:
on ubuntu 14.04, expect-dev does not exist and
unbuffer is provided by expct


gurneyalex:
my bad

PR #506, opened 2017-10-30
Revert "[FIX] travis_transifex: Make the check before"

moylop260:
This reverts commit 27792ba2e1e95212c1a0670784bdd3cd1a5b33c0.

Because we have red travis result and it was a direct commit instead of use PR

This PR is just to test if there is a real red after this commit


moylop260:
Closing since that https://github.com/OCA/maintainer-quality-tools/pull/504 is fixed

PR #512, opened 2017-11-10
[FIX] test_server: Use configparser py2 and py3 compatibility

moylop260:
Fix the error `No module named configparser` for py2 instances where is not installed the package `configparser`


moylop260:
Fast-tracking since that other builds 10.0 is failing...

PR #518, opened 2018-01-09
[UPD] .gitignore

mreficent:
Updating the .gitignore as https://github.com/OCA/maintainer-tools/blob/master/.gitignore, requested by @lasley in https://github.com/OCA/maintainer-tools/pull/318.


sample_files/.gitignore
```diff
@@ -36,10 +40,23 @@ coverage.xml
...
+*.pot
 
 # Pycharm
 .idea
 
+# Eclipse
+.settings
+
+# Visual Studio cache/options directory
+.vs/
```
pedrobaeza: Shouldn't this be .vscode?

mreficent: https://github.com/OCA/maintainer-tools/pull/316

pedrobaeza: Yeah, but now see what my colleague @Yajo added in other PR: https://github.com/OCA/maintainer-quality-tools/pull/519/files#diff-b1d6df5420097c477a0feb225b280b64R42

lasley: Looks like both are valid. One is project folder for Visual Studio, the other is Visual Studio Code. Ambiguous naming, but different things.

Can we add both please?

mreficent: Done

sample_files/.gitignore
```diff
@@ -55,5 +73,5 @@ docs/_build/
 *~
 *.swp
 
-# OSX Files
-*.DS_Store
+# OCA rules
+!static/lib/
```
lasley: Whoah whoah whoah whoah whoah. This is a thing?!?!?!?!

Welp TIL

pedrobaeza: What this means?

lasley: Gitignore supports negation patterns, which is what this is. In web repos, we have a valid `static/lib` folder that is typically excluded by the [`lib/` exclusion for Python distribution](https://github.com/Eficent/maintainer-quality-tools/blob/cbc551372084c82ab81cdb0b3cea806e79115f9f/sample_files/.gitignore#L19). Adding this negation after the linked declaration allows for both to exist in harmony.

If you take a look at the [`OCA/web` repo gitignore](https://github.com/OCA/web/blob/11.0/.gitignore), you'll notice we had to actually remove the `lib/` declaration. @mreficent suggestion fixes that.

So yeah, that's my TIL (today I learned) 🚀

pedrobaeza: Indeed!

mreficent:
Why v11/Python 3.5 check of travis fails?

PR #522, opened 2018-01-11
[IMP] Remove pylint beta messages from PR cfg file

Yajo:
This way we DRY the files and also allow the PR cfg file to be used in standalone mode.

@Tecnativa

### DIFF
--- a/travis/cfg/travis_run_pylint_pr.cfg
+++ b/travis/cfg/travis_run_pylint_pr.cfg
@@ -19,92 +19,21 @@ license_allowed=AGPL-3,GPL-2,GPL-2 or any later version,GPL-3,GPL-3 or any later
 disable=all
 
 # Enable message and code:
-#   api-one-deprecated - W8104
-#   api-one-multi-together - W8101
-#   attribute-deprecated - W8105
-#   class-camelcase - C8104
-#   create-user-wo-reset-password - W7905
-#   copy-wo-api-one - W8102
-#   dangerous-filter-wo-user - W7901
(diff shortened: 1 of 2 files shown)


pedrobaeza:
Travis is failing

Yajo:
Let's see now.

Yajo:
Yep, 💚

pedrobaeza:
I understand that no check is removed. You only concatenate both - PR and normal -, isn't it?

Yajo:
Actually normal + beta, or pr + beta if I'm not wrong. @moylop260 should know it better.

moylop260:
Hello @Yajo 

normal.cfg + pr.cfg - beta messages for PRs
normal.cfg - beta messages for stable branches

I mean, beta messages are not considered for affecting the status build result from final result.

Example:
normal.cfg enables:
 - check1
 - check2

pr.cfg enables:
 - check3
 - check4

beta.cfg enables:
  - check1
  - check4

If a PR or stable branch emit the messages `check1` or/and `check4` then will be result green because are beta messages (these checks are not approval 100% yet)

If a PR or stable branch emit the message `check2` and/or `check3` then will be result red because are not silent from beta message.

Note: If you have a better solution for this feature is welcome.
Thanks!

Yajo:
AFAIK the better solution is this PR, isn't it?

What it does is unduplicate checks from beta cfg file, and add them manually through CLI.

This way, behavior should remain the same while PR cfg can be used in standalone mode (by not including beta checks)

PR #525, opened 2018-01-12
[FIX] tests: Update dictionary of expected errors.

moylop260:
[FIX] .travis.yml: Fix false red for stable branches, forcing PR flag
[FIX] self_tests: Returns exit code with the result of the tests

Currently there is a error with the dictionary expected but is not visible because show a false green.
Other issue is that the travis result of the branch stable is red because the flag of PR was off.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -51,8 +51,8 @@ install:
   - git --git-dir=${TRAVIS_BUILD_DIR}/.git add --all  # All modules moved are modules changed to test PR changes
 
 script:
-  - coverage run --append ./travis/travis_run_tests 8.0  # only used if VERSION not set in env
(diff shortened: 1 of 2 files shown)


Yajo:
If we're going to use the pull request mode always, then why don't we just remove the other mode and forget about this env variable?

moylop260:
@Yajo @lasley 
The MQT's PRs are green because the expected errors changes if `TRAVIS_PULL_REQUEST` is enabled and use the `pylint_pr.cfg` but if the branch is merged this flag is disabled from travis [more info about flag](https://docs.travis-ci.com/user/environment-variables/#Default-Environment-Variables) then show error:
 - https://travis-ci.org/OCA/maintainer-quality-tools/jobs/328154993#L699

We need enable this flag to run lints for MQT code including pylint_pr.cfg checks.

PR #526, opened 2018-01-22
[FIX][travis] 'by_msg' not in errors

mreficent:
This PR fixes:

    Traceback (most recent call last):
      File "/home/travis/maintainer-quality-tools/travis/test_pylint", line 59, in <module>
        result = run_pylint.pylint_run(is_pull_request, version, git_work_dir)
      File "/home/travis/maintainer-quality-tools/travis/run_pylint.py", line 170, in pylint_run
        (key, value) for key, value in real_errors['by_msg'].items()
    KeyError: 'by_msg'

It happens when you delete all modules in a repository, thus there aren't error messages because there is no module.

### DIFF
--- a/travis/run_pylint.py
+++ b/travis/run_pylint.py
@@ -167,8 +167,8 @@ def pylint_run(is_pr, version, dir):
 
     real_errors = main(cmd, standalone_mode=False)
     res = dict(
-        (key, value) for key, value in real_errors['by_msg'].items()
-        if key not in beta_msgs)
+        (key, value) for key, value in (real_errors.get(
+            'by_msg') or {}).items() if key not in beta_msgs)
     count_errors = get_count_fails(real_errors, list(beta_msgs))
     count_info = "count_errors %s" % count_errors
     print (count_info)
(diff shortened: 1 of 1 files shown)


travis/run_pylint.py
```diff
@@ -168,7 +168,7 @@ def pylint_run(is_pr, version, dir):
     real_errors = main(cmd, standalone_mode=False)
     res = dict(
         (key, value) for key, value in real_errors['by_msg'].items()
-        if key not in beta_msgs)
+        if key not in beta_msgs) if 'by_msg' in real_errors else dict()
```
pedrobaeza: I think it's easier and more readable this way:
```python
    res = dict(
         (key, value) for key, value in real_errors.get('by_msg', {}).items()
        if key not in beta_msgs)
```

moylop260: beteer use (real_error.get(‘by_msg’) or {}) incorder to avoid the same error if the key is defined with None value

mreficent: The problem is the <code>.items()</code>. You can't do <code>{}.items()</code>

pedrobaeza: Of course you can do it. It returns `[]`.

mreficent: Done!

PR #527, opened 2018-01-23
[FIX] runbot builds on 11.0

gurneyalex:
in v11, phonenumbers is an undeclard optional dependency. Not having this
module will trigger a warning in the logs and then a failed runbot
build.

### DIFF
--- a/travis/travis_install_nightly
+++ b/travis/travis_install_nightly
@@ -112,6 +112,10 @@ sed -i 's/^requests\=\=/requests[security]\=\=/g' ${ODOO_PATH}/requirements.txt
 
 pip install -q --no-binary pycparser -r ${ODOO_PATH}/requirements.txt
 pip install -q QUnitSuite coveralls codecov
+# in v11, phonenumbers is an undeclared optional dependency. Not having this
+# module will trigger a warning in the logs and then a failed runbot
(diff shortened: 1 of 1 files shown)


pedrobaeza:
This is not going to work for runbot, as it doesn't use this stack.

gurneyalex:
I thought travis2docker was parsing travis.yml and executing stuff??

gurneyalex:
See also https://github.com/OCA/OCB/pull/696

pedrobaeza:
@moylop260 can you please confirm what @gurneyalex said? It's phonenumbers going to be available on runbot builds with this?

gurneyalex:
even if https://github.com/OCA/OCB/pull/696 is merged (and I think it should be because otherwise it is painful to remember to install the module for my own customers), having this here does not do harm as it allows to have the module available in the travis tests run on odoo/odoo. You get a noop on OCA/OCB as the module will already be installed.

pedrobaeza:
Yeah, of course, although Travis MQT doesn't consider warning as errors, but no problem for having it.

moylop260:
FYI let's to see that odoo tell us:
https://github.com/odoo/odoo/pull/22459

PR #531, opened 2018-03-12
[FIX] adapt behavior to docs

hbrunn:
in order to get the version autodetection the docs claim at the end of https://github.com/hbrunn/maintainer-quality-tools/blob/be6e4222c98128950c61bfad1b1ec64706f86a44/git/README.md, we need to pass an empty string to https://github.com/hbrunn/maintainer-quality-tools/blob/be6e4222c98128950c61bfad1b1ec64706f86a44/travis/run_pylint.py#L124

I don't really see the point of checking against the empty string there, maybe we should just check for falsyness?

### DIFF
--- a/travis/run_pylint.py
+++ b/travis/run_pylint.py
@@ -121,7 +121,7 @@ def get_modules_cmd(dir):
 
 
 def version_validate(version, dir):
-    if version == "" and dir:
+    if not version and dir:
         repo_path = os.path.join(dir, '.git')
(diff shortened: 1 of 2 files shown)


hbrunn:
@moylop260 thanks for merging, but i'm still curious about what i ask above - why can't we just check for falsyness in the snippet linked there?

hbrunn:
ah, you did that then, nice

PR #536, opened 2018-04-05
[ADD] gengo library in requirements

mreficent:
`base_gengo` module uses this library and Odoo doesn't have put in their requirements file (yet).


pedrobaeza:
This should be in OCB instead.

Issue #539, opened 2018-04-14
Strange problem in Transifex + odoo/odoo builds in Travis


pedrobaeza:
I'm experimenting a strange problem in both stages in several repositories. Examples:

https://travis-ci.org/OCA/bank-payment/builds/366461895
https://travis-ci.org/OCA/l10n-spain/builds/366268879

It coincides in time more or less with the arrival of #537. Is there any relation?

@moylop260 @StefanRijnhart @mreficent

sbidoul:
My bet is that github has had a timezone bug when preparing the current tar.gz for 10.0. It does not seem to be related to #537

pedrobaeza:
But the date seems correct...
```
time stamp 2018-04-14 14:29:07
```
and it's only happening in that build, not in OCA/OCB one. As said, very strange...

pedrobaeza:
It's only happening on 10.0 build, but not the rest (I have seen 9.0 and 11.0 successful builds). Maybe something incorrect in GitHub tar? How is it obtained?

pedrobaeza:
Thanks for checking. Then, I can close this.

pedrobaeza:
What are the advantages of using that method?

moylop260:
Advantages using git clone instead of tar file:
 - Running MQT script from your local machine or docker you can re-user the odoo repository to upgrade it (`git pull`)
 - Using `git depth=10` like as travis-ci uses we will have a faster download.
 - We can know the odoo sha of the build if there is a error you can reproduce it using the same sha.
 - We can verify a custom change applied to dooo repository if there are a `sed` executed or `patch` using `git diff` and return to original state.

And the other one, remove this flaky error using tar files.

pedrobaeza:
Well, let's go with that then! You're the master of MQT, so you know where to touch.

moylop260:
FYI https://github.com/OCA/maintainer-quality-tools/pull/540

PR #540, opened 2018-04-16
[REF] install: Use git instead of tar to download odoo

moylop260:
* Advantages using git clone instead of the tar file:
 - Running MQT script from your local machine or docker you can re-use the odoo repository to upgrade it (`git pull`)
 - Using `git depth=50` like as travis-ci uses we will have a faster download.
 - We can know the odoo sha of the build if there is an error you can reproduce it using the same sha.
 - We can verify a custom change applied to odoo repository if there is a `sed` executed or `patch` using `git diff` and return to original state.
 - Remove flaky error using tar files:
    - https://github.com/OCA/maintainer-quality-tools/issues/539
 - We can use a odoo's PR to test it

More info about: 
https://github.com/OCA/maintainer-quality-tools/issues/539#issuecomment-381648037

- [x] TODO: Create a dummy PR using this repository
  - Waiting green result https://github.com/OCA/l10n-spain/pull/829

### DIFF
--- a/travis/travis_install_nightly
+++ b/travis/travis_install_nightly
@@ -72,28 +72,24 @@ if [[ $BRANCH == *"/"*  ]]; then
     export BRANCH=${BRANCH/\//-}
 fi
 export ODOO_PATH=${HOME}/$REPO_NAME-$ODOO_BRANCH
-if [ -z "${REPO_CACHED}"  ]; then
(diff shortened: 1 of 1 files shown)


moylop260:
@pedrobaeza 
It's done!

pedrobaeza:
I think we can go on with this technical PR, as no more reviews are expected.

PR #541, opened 2018-04-17
[IMP] test_server: update test modules instead if they were preinstalled

mreficent:
Check which modules are preinstalled. If some test modules were preinstalled, then they will be updated in the tests instead of installed.

### DIFF
--- a/travis/test_server.py
+++ b/travis/test_server.py
@@ -101,6 +101,96 @@ def has_test_errors(fname, dbname, odoo_version, check_loaded=True):
     return len(errors)
 
 
+def installed_modules(fname, dbname, preinstall_modules=None):
+    """
+    Check a list of log lines for modules installed.
+    """
+    # Rules defining checks to perform
+    # this can be
+    # - a string which will be checked in a simple substring match
+    # - a regex object that will be matched against the whole message
+    # - a callable that receives a dictionary of the form
+    #     {
+    #         'loglevel': ...,
+    #         'message': ....,
+    #     }
+    if preinstall_modules is None:
+        preinstall_modules = ['base']
+    infos_ignore = [
+        'Computing parent',
+        'Storing computed',
+        'deleted',
+        'skip',
+        'with ID',
+        ' /',
+        ' modules',
+        ]
+    infos_report = [
+        lambda x: x['loglevel'] == 'INFO',
+        ]
+
+    def make_pattern_list_callable(pattern_list):
+        for i in range(len(pattern_list)):
+            if isinstance(pattern_list[i], string_types):
+                regex = re.compile(pattern_list[i])
+                pattern_list[i] = lambda x, regex=regex:\
+                    regex.search(x['message'])
+            elif hasattr(pattern_list[i], 'match'):
+                regex = pattern_list[i]
+                pattern_list[i] = lambda x, regex=regex:\
+                    regex.search(x['message'])
+
+    make_pattern_list_callable(infos_ignore)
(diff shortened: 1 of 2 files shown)


StefanRijnhart:
Elegant or not, you have not convinced me with this code ;-) I am still in favour of simply including l10n_generic_coa as an 'application' instead.

mreficent:
But probably the problem is not just with l10n_generic_coa. We don't know. After an exception of l10n_generic_coa, may we need more exceptions for modules like l10n_us, that also is preinstalled. IMO, it doesn't make sense to install any module with `--init` two times and we should not allow that. The tests have to make sense, and if any module should be installed two times, the second time should be an upgrade or maybe we can uninstall it to reinstall it again.

StefanRijnhart:
We don't have to take my gut feeling for it. I tested it. The following (without including l10n_generic_coa in the localization modules) runs fine:
```
bin/start_odoo -d 90test -i account_cancel,hr_contract,base_iban,point_of_sale,account_accountant,base_action_rule,hr_payroll,barcodes,web,base_setup,procurement,web_tip,analytic,mail,sale_stock,stock,product,purchase,hr,bus,report_intrastat,web_calendar,delivery,report,hr_payroll_account,decimal_precision,hr_expense,account,resource,base_vat,web_planner,sale,web_kanban,hr_holidays,calendar,sales_team,stock_account
bin/start_odoo -d 90test -i l10n_ae,l10n_ar,l10n_at,l10n_au,l10n_be,l10n_be_hr_payroll,l10n_be_hr_payroll_account,l10n_be_intrastat,l10n_be_invoice_bba,l10n_bo,l10n_br,l10n_ca,l10n_ch,l10n_cl,l10n_cn,l10n_cn_small_business,l10n_cn_standard,l10n_co,l10n_cr,l10n_de,l10n_de_skr03,l10n_de_skr04,l10n_do,l10n_ec,l10n_es,l10n_et,l10n_eu_service,l10n_fr,l10n_fr_certification,l10n_fr_fec,l10n_fr_hr_payroll,l10n_fr_pos_cert,l10n_fr_sale_closing,l10n_gr,l10n_gt,l10n_hn,l10n_hr,l10n_hu,l10n_in,l10n_in_hr_payroll,l10n_in_schedule6,l10n_it,l10n_jp,l10n_lu,l10n_ma,l10n_multilang,l10n_mx,l10n_nl,l10n_no,l10n_nz,l10n_pa,l10n_pe,l10n_pl,l10n_pt,l10n_ro,l10n_sa,l10n_sg,l10n_si,l10n_syscohada,l10n_th,l10n_tr,l10n_uk,l10n_us,l10n_uy,l10n_ve,l10n_vn --test-enable --stop-after-init
```

mreficent:
But you are not testing l10n_generic_coa. With my PR, it automatically does:

```
bin/start_odoo -d 90test -i l10n_ae,l10n_ar, ..., l10n_vn -u l10n_generic_coa --test-enable --stop-after-init
```

and this way l10n_generic_coa also is tested.

StefanRijnhart:
l10n_generic_coa is installed (and tested but for the fact that it does not contain any tests) in the application run automatically already as per account's post_init_hook https://github.com/odoo/odoo/blob/9.0/addons/account/__init__.py#L25

moylop260:
Is this PR valid yet after https://github.com/OCA/OCB/pull/730?

mreficent:
No, but I will make a new PR to correct a few things this PR contained.

PR #543, opened 2018-04-18
[FIX] Exclude l10n_generic_coa from modules

mreficent:
Solves https://github.com/OCA/OCB/pull/730.


StefanRijnhart:
I would put the exception in `get_localizations_with_dependents` but this does the trick as well

mreficent:
I didn't want to exclude the dependents of l10n_generic_coa, so I went with the safest way 😅

PR #544, opened 2018-04-20
[FIX] Test_server: minor touches

mreficent:
Some minor touches.

### DIFF
--- a/travis/test_server.py
+++ b/travis/test_server.py
@@ -214,11 +214,13 @@ def setup_server(db, odoo_unittest, tested_addons, server_path, script_name,
     if the database template exists then will be used.
     :param db: Template database name
     :param odoo_unittest: Boolean for unit test (travis parameter)
-    :param tested_addons: List of modules that need to be installed
+    :param tested_addons: (list) Modules that need to be installed
     :param server_path: Server path
-    :param travis_build_dir: path to the modules to be tested
+    :param script_name: name of the main server file
     :param addons_path: Addons path
     :param install_options: Install options (travis parameter)
+    :param preinstall_modules: (list) Modules that should be preinstalled
+    :param unbuffer: keeps output colors
     :param server_options: (list) Add these flags to the Odoo server init
     """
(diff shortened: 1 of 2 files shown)


travis/test_server.py
```diff
@@ -231,7 +233,6 @@ def setup_server(db, odoo_unittest, tested_addons, server_path, script_name,
     except subprocess.CalledProcessError:
         print("Using previous openerp_template database.")
     else:
-        # unbuffer keeps output colors
```
moylop260: Should we remove this comment?

mreficent: It's not removed, just moved to https://github.com/OCA/maintainer-quality-tools/pull/544/files#diff-4b42f344199e5694ffee3d2a57dd80c7R223

travis/test_server.py
```diff
@@ -352,15 +353,15 @@ def main(argv=None):
         print("WARNING!\nNothing to test- exiting early.")
         return 0
     else:
-        print("Modules to test: %s" % tested_addons)
-    # setup the base module without running the tests
+        print("Modules to test: %s" % tested_addons_list)
```
pedrobaeza: Don't change this to the list, but put the comma separated variable (the previous one).

mreficent:
@pedrobaeza The unique signature change is in `setup_server` method, where the `addons` variable was an unused string and I changed it to a list variable because lists are more usable than strings and because in the docstring already said it's a list.

travis/test_server.py
mreficent: Well then, I will update the PR by putting all outputs to comma separated string format. But I prefer the list output because :laughing:

pedrobaeza: OK, then keep it. I don't have a strong opinion for forcing you to change the other.

pedrobaeza:
OK also about the "fake" signature change

travis/test_server.py
mreficent: 😂 thaks

Issue #545, opened 2018-04-23
Switch to W504 from W503 (line break before a logical operator)


yelizariev:
Could OCA use W504, i.e.
```
if (vals.get('model') == 'ir.ui.view'
        and not self.env.context.get('duplicate_view_for_website')):
```

instead of W503, i.e.
```
if (vals.get('model') == 'ir.ui.view' and
        not self.env.context.get('duplicate_view_for_website')):
```
See
https://github.com/PyCQA/pycodestyle/issues/498
https://github.com/PyCQA/pycodestyle/pull/502/files

Feel free to add :+1: or :-1: to this post

legalsylvain:
I personaly against such rule changes.
 
* It is matter of tastes, and doesn't provides any improvment.
* requires a lot of changes, travis will fail if we apply this change of rule.
* maintainers have other things more important than cosmetics changes like this. (IMO)

To be clear, I'm not more in favor of W504 than W503 or whatever, I just think that changing that rule will provides problems, without any improvment.

Regards.

yelizariev:
There is always a way to apply changes softly (deactivate W503 and make W504 beta).

> It is matter of tastes, and doesn't provides any improvment.

It's actually allows read code easily. See

[1] https://mail.python.org/pipermail/python-ideas/2016-April/039774.html
[2] http://bugs.python.org/issue26763
[3] https://hg.python.org/peps/rev/3857909d7956

> maintainers have other things more important than cosmetics changes like this. (IMO)

The contributors too. Why OCA force them to update they code to W503?

StefanRijnhart:
Agree with @legalsylvain to allow both styles.

yelizariev:
Allowing both is also a solution

moylop260:
Changing this style rule we will have too many travis red results.

If the current style was changed and the new one will make us headaches... then what about remove this rule?

Adding it to ignore section of the flake8 configuration file: W503

yelizariev:
Also, we can disable W503 and apply W504 since odoo 12

moylop260:
> Allowing both is also a solution

The technical solution is ignoring both from the configuration file of flake8.

FYI we have advanced script for pylint (to choose a special configuration file for each odoo version or PR or betas...) but for flake8 (this check is a flake8 one) we don't have this options...

moylop260:
FYI I have taken the liberty to have created the following PR:
 - https://github.com/OCA/maintainer-quality-tools/pull/546

[...]

PR #546, opened 2018-04-23
[REF] flake8.cfg: Allow W504 and W503 flake8 style

moylop260:
More info about https://github.com/OCA/maintainer-quality-tools/issues/545

### DIFF
--- a/travis/cfg/travis_run_flake8.cfg
+++ b/travis/cfg/travis_run_flake8.cfg
@@ -3,6 +3,6 @@
 # F811 is legal in odoo 8 when we implement 2 interfaces for a method
 # F999 pylint support this case with expected tests
 # W503 changed by W504 and OCA prefers allow both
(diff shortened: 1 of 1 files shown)


travis/cfg/travis_run_flake8.cfg
```diff
@@ -2,6 +2,7 @@
 # E123,E133,E226,E241,E242 are ignored by default by pep8 and flake8
 # F811 is legal in odoo 8 when we implement 2 interfaces for a method
 # F999 pylint support this case with expected tests
-ignore = E123,E133,E226,E241,E242,F811,F601
+# W503 changed by W504 and OCA prefers allow both
+ignore = E123,E133,E226,E241,E242,F811,F601,W503
```
pedrobaeza: Shouldn't we add both and not rely on defaults?

PR #538, opened 2018-04-11
[IMP] use higher python version, disable system site packages

hbrunn:
see discussion in https://github.com/OCA/web/pull/909#issuecomment-380045633 and https://github.com/OCA/web/pull/889#issuecomment-379801321

Issue #432, opened 2017-02-21
Phantomjs test timeout in travis


StefanRijnhart:
Hi, this test[1] succeeds locally, but times out with the OCA travis config[2]. Is this a known issue?

```
        code = """
            setTimeout(function () {
                if (!$('.oe_list_button_import').is(":visible")) {
                    throw new Error('Import button is not accessible.');
                }
                console.log('ok');
            }, 3000);
        """
        link = '/web#action=%s' % self.env.ref('base.action_res_users').id
        with mute_logger('openerp.addons.base.res.res_users'):
            # Mute debug log about failing row lock upon user login
            self.phantom_js(
                link, code, "$('button.oe_list_add')",
                login=self.env.user.login)
```

[1] https://github.com/OCA/server-tools/pull/743/files#diff-42657ea13cced82fd0efda56ecf8fef0R26
[2] https://travis-ci.org/OCA/server-tools/jobs/203880321

moylop260:
Locally what commands are you executing?

moylop260:
Forge it, I can reproduce it using original travis image with travis2docker.
FYI I can't reproduce it using the image used from runbot.

Maybe is a issue with library versions:
travis image where is reproduce it:
`phantomjs --version`
`1.9.8`

vx image where is not reproduce it:
`phantomjs --version`
`1.9.0`

What version do you have locally?

StefanRijnhart:
That's it. It works fine on 1.9.7. I have another machine with 2.1.1, and that version does not stop execution upon 'throw Error' so it times out instead of falsifying the behaviour. Another difference is that my waiting condition _$('button.oe_list_add')_ suddenly evaluates to true before it is loaded and I needed to switch to  _$('button.oe_list_add').**length**_.

Thanks for your help!

StefanRijnhart:
Still not working. It works locally with phantomjs 2.1.1 and
```
bin/start_odoo  -d test_base_import_security_group -i base_import_security_group --test-enable --stop-after-init --addons-path=../odoo/addons,../server-tools --workers=0
```
(config file is empty)

While on a travis config with phantomjs 2.1.1[1] I'm still getting the error timeout[2]

[1] https://travis-ci.org/OCA/server-tools/jobs/203995068#L272
[2] https://travis-ci.org/OCA/server-tools/jobs/203995068#L1890

moylop260:
The real issue is:
 - A test using a external connection (phantomjs) with previous record modified.
[...]

PR #548, opened 2018-05-03
[REF] sample_files: .travis.yml - Higher postgresql version

moylop260:
* By default postgresql-9.1 is installed but there is issue related:
    - https://github.com/OCA/maintainer-quality-tools/issues/432#issuecomment-281580935
 * Better use higher PostgreSQL version

Continuation from https://github.com/OCA/maintainer-quality-tools/pull/538

### DIFF
--- a/sample_files/.travis.yml
+++ b/sample_files/.travis.yml
@@ -9,6 +9,10 @@ python:
   - "2.7.13"
 
 addons:
+  # By default postgresql-9.1 is installed but there is issue related:
+  #  https://github.com/OCA/maintainer-quality-tools/issues/432#issuecomment-281580935
+  # Better use higher PostgreSQL version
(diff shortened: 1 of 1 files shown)


sample_files/.travis.yml
```diff
@@ -9,6 +9,10 @@ python:
   - "2.7.13"
 
 addons:
+  # By default postgresql-9.1 is installed but there is issue related:
+  #  https://github.com/OCA/maintainer-quality-tools/issues/432#issuecomment-281580935
+  # Better use higher PostgreSQL version
+  postgresql: "9.4"
```
simahawk: why not 9.5?

moylop260: 9.5 is fine too and It is the official version for `Ubuntu-16.04` then changed.

PR #550, opened 2018-05-08
Using TRAVIS_BUILD_DIR variable instead of hardcoded path

PCatinean:
When using MQT outside of travis the hardcoded /build/ dir of Travis is causing an issue by looking in the wrong place.

This offers a more flexible way to check the code that needs to be tested because even in gitlab we can define the TRAVIS_BUILD_DIR.

Downside is that if someone places other pieces of code in the /build/ dir next to the project they will be ignored (if that was ever the plan).

A better way would be to use a separate env var and handle default values but that might introduce new dependencies (such as envsusbt and such)

Close https://github.com/OCA/l10n-spain/pull/848

### DIFF
--- a/cfg/.coveragerc
+++ b/cfg/.coveragerc
@@ -2,7 +2,7 @@
 
 [report]
 include =
-    */build/*
+    ${TRAVIS_BUILD_DIR}/*
 
 omit =
     */scenario/*


pedrobaeza:
@moylop260 is that variable correct to be put here in this file?

PCatinean:
I've tested it with gitlab and it resolved well and pulls the proper code coverage. I assume we also need to wait for the automated tests to finish and also if it's important that some other code in the build dir (possibly manually being cloned as a step) is also something someone depends on.

cfg/.coveragerc
```diff
@@ -2,7 +2,7 @@
 
 [report]
 include =
-    */build/*
+    *$TRAVIS_BUILD_DIR*
```
moylop260: Use `/`

PCatinean: At the very end? (since it has a trailing slash due to absolute path)

moylop260:
FYI I have created the following dummy PR in order to test it:
 - https://github.com/OCA/l10n-spain/pull/848

PCatinean:
@moylop260 my pleasure. Would it be a good idea to also include something like $CODECOV_OMIT in the omit section of .coveragerc while we're at it and give also this option?

PCatinean:
@moylop260 changes applied, test on gitlab worked great using the new variables.

PCatinean:
@moylop260 squashed everything in one commit and removed the one you mentioned. I would like to create and get the other PR in as well because otherwise you are forced to have a test-coverage of all modules including dependencies which you might not want to manage yourself (or at least at certain stages).

moylop260:
Currently, we have an INCLUDE and EXCLUDE environment variables... but we need to test if a separated by commas is a valid value...
If it is not then we will need to parse the original value and change the coveragerc based on it...

I mean, I agree to create this change for other PR with a depth review. 

Thanks!

PCatinean:
Ah so you are saying to use the existing exclude env vars (applying for tests) from the coverage as well automatically?

PR #554, opened 2018-05-29
make rst syntax errors blocking checks

### LINKED ISSUE #553: rst syntax errors must become blocking issues
pypi.org now prevents uploading packages with incorrect rst descriptions.

The README.rst check must therefore become a blocking error.

Additionally, it would be great if the rst check would be applied to `readme/*.rst` too.

For reference, pypi.org uses [readme_renderer](https://pypi.org/project/readme_renderer/) for validation.

sbidoul:
Closes #553

### DIFF
--- a/travis/cfg/travis_run_pylint.cfg
+++ b/travis/cfg/travis_run_pylint.cfg
@@ -19,6 +19,7 @@ disable=all
 #   reimported - W0404
 #   relative-import - W0403
 #   return-in-init - E0101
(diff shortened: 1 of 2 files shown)


sbidoul:
Need guidance to update the tests.

pedrobaeza:
You have to raise the number of expected errors https://github.com/OCA/maintainer-quality-tools/blob/master/.travis.yml#L38 (1 more as far as I see). And I think with that it's enough.

PR #560, opened 2018-07-05
[FIX] install: Use compatible less version for odoo

moylop260:
More info about: https://github.com/odoo/odoo/issues/25632
Note: It is not a good idea pinned the version but currently we will have all our runbot's instances without connection

UPDATED:
You can see an example of the error connecting a recent build from our runbot:
 - https://runbot.odoo-community.org/runbot/226/11.0

### DIFF
--- a/travis/travis_install_nightly
+++ b/travis/travis_install_nightly
@@ -28,6 +28,7 @@ fi
 
 # Install less
 ln -s `which nodejs` $HOME/maintainer-quality-tools/travis/node
+# Remove fixed version after fix https://github.com/odoo/odoo/issues/25632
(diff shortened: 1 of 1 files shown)


moylop260:
commit message comment or pr comment?

moylop260:
“Please, revert this PR after fix”

pedrobaeza:
I was thinking on the code itself on `travis_install_nightly`, putting the link to the Odoo issue, so someone revisiting that code, can check the issue and decide if we can remove it.

pedrobaeza:
Let's merge it for fixing the problem. Let's see if we can fix it upstream soon.

PR #562, opened 2018-07-20
WIP: [REF] .travis.yml: Use python 3.7

moylop260:
https://docs.travis-ci.com/user/languages/python/#Specifying-Python-versions


.travis.yml
```diff
@@ -40,6 +40,10 @@ matrix:
       env:
         VERSION="11.0" ODOO_REPO="OCA/OCB" TESTS="1" LINT_CHECK="0"
         INCLUDE="test_module,second_module"
+    - python: 3.7-dev
```
Yajo: Why dev?

moylop260: The list of the documentation has dev.
I think that is because they don’t have updated the image base with the recently released stable yet.

https://docs.travis-ci.com/user/languages/python/#specifying-python-versions

Yajo: I guess those are just examples. Have you actually tested it without the -dev?

moylop260:
@yajo
Could you review the build result?
Tell me if we need something else
regards

PR #563, opened 2018-07-24
[UPD] Cleanup gitlab-ci file using python-3.5 image

RoelAdriaans:
This is, imho, a cleaner gitlab ci example.
Uses the docker python 3.5 image with an extra service dependency on the postgresql image. 

Testing can be done on the public shared gitlab runners, or private runners when needed.

An example where a public oca repository is tested can be found https://gitlab.com/road-support/oca-account-invoicing/
(It's currently failing because the codecoverage has gone down, test and lint check are green)

### DIFF
--- a/sample_files/.gitlab-ci-2.yml
+++ b/sample_files/.gitlab-ci-2.yml
@@ -1,0 +1,66 @@
+image: python:3.5
+stages:
+  - test
+
+services:
+  - postgres:10
+
+variables:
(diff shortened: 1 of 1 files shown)


moylop260:
I think that both ways are valid.
Could you create another file for this second way, please?

RoelAdriaans:
@moylop260 No problem, renamed the file..

PR #566, opened 2018-08-20
[FIX] Include six in requirements.txt

mreficent:
It happens that `travis/git_run.py` and `travis/test_server.p`y use `six` library, but that library is not in `requirements.txt` file. Thus, this PR fixes that.

Also, as a bonus, I updated the disable message commented sections.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -4,7 +4,7 @@ pyopenssl>=16.2.0
 pyyaml
 coveralls
 unittest2
-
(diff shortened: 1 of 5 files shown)


mreficent:
@moylop260

moylop260:
It was merged from pylint-odoo
It os required yet?

mreficent:
Of course.

PR #567, opened 2018-08-22
[REF] cfg: Rename param manifest_required_author

luisg123v:
Since it was renamed on OCA/pylint-odoo@8dcb1eb9df1e, from
`manifest_required_author` to `manifest_required_authors` (from singular
to plural), it needs to be renamed here, too.


luisg123v:
Thanks @pedrobaeza 

@moylop260  could you approbe this PR, please?

PR #570, opened 2018-08-23
[FIX] self_tests: Adjust expected values after pylint change

luisg123v:
In a recent change on pylint-odoo [1], it was introduced a fix to
recognize a badly-formatted version in manifest, which now depends on
the Odoo version. When the version is provided and it's a numeric
version value (e.g. not "master") it's taken into account to determine
whether the version defined in the manifest is formatted correctly.

Since versions in tested modules are as follows:
- Two set to 8.0
- Four set to wrong values (e.g. 1.0)

When the version is numeric but it's not 8.0, there will be six errors,
otherwise there will be four. In these tests, the only case when it's
numeric but it's not 8.0, is when testing 7.0.

[1] Pylint change: https://github.com/OCA/pylint-odoo/commit/6b1f1e1bfaf

### DIFF
--- a/travis/self_tests
+++ b/travis/self_tests
@@ -90,6 +90,7 @@ class MainTest(unittest.TestCase):
         self.modules_cmd = run_pylint.get_modules_cmd(self.git_work_dir)
         self.beta_msgs = run_pylint.get_beta_msgs()
         self.errors_dict = EXPECTED_ERRORS.copy()
+        self.maxDiff = None
 
     def test_modules_to_preinstall(self):
         res = [x for x in self.to_preinstall if x in self.addons_list]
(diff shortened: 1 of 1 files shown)


luisg123v:
@pedrobaeza  
Take into account the fix was merged recently. Before that, the lint was not taking into account the Odoo version.

pedrobaeza:
Ah, OK

luisg123v:
@pedrobaeza 
Just to be sure, I created a dummy PR on https://github.com/OCA/server-tools/pull/1349  to revert that change, to see if the lint is triggered now, and it is:

```
scheduler_error_mailer/__manifest__.py:6: [C8106(manifest-version-format), ]
Wrong Version Format "10.0.1.0.0" in manifest file. Regex to match: "(11\.0)\.\d+\.\d+\.\d+$"
```

([source](https://travis-ci.org/OCA/server-tools/jobs/419835653#L662))

BTW, could you approve this change, please? master is currently red because of this

CC @moylop260

luisg123v:
@pedrobaeza you're welcome, I'm happy to help and to contribute to the community, and I will continue doing so as much as I'm able to.

PR #283, opened 2015-10-18
Add repo dependency installation self tests

dreispt:

PR #574, opened 2018-09-12
[IMP] test_server: Patch odoo to create unlogged DB tables (#283)

luisg123v:
This commit causes Odoo to be patched so all database tables are created
as unlogged.

This is applied only for testing instances because, if a docker
container is stopped without firstly saving the changes, all records
will be deleted; so it's not suitable for working under other
environments, e.g. locally.

This is compatible with the versions 7.0 to master (12.0).


luisg123v:
This will be addressed on https://github.com/OCA/maintainer-quality-tools/pull/573, hence I'm closing this.

PR #576, opened 2018-09-28
[FIX] test_server: Ignore trailing whitespaces on log messages

luisg123v:
Odoo recently introduced a change in the way logs are printed, adding
performance information when logging HTTP requests:
https://github.com/odoo/odoo/commit/c44fe912329e2d4c7541ed810526d4a069f930c4#diff-d301b3dd399c26a7f38fb31c9e1dfbddR136

However, that causes all messages who don't containg such information to
contain a trailing whitespace, which causes the message "Modules
loaded." to not be found.

This commit improves the regexp used to match messages in logs, to not
take into account trailing whitespaces.


luisg123v:
CC @moylop260

luisg123v:
I just noticed the message "Modules loaded." is detected anyway, thanks to https://github.com/oca/maintainer-quality-tools/commit/be28ef3e, so I edited the description accordingly.

PR #577, opened 2018-09-28
[IMP] test_server: Improve error msg when "Modules loaded." is not found

luisg123v:
This improves the error message, so it's clearer.


luisg123v:
CC @moylop260

PR #582, opened 2018-11-23
[REF] pylint: Manifest key 'development_status' required

moylop260:
Part of https://github.com/OCA/pylint-odoo/issues/199


pedrobaeza:
It shouldn't be requires, as it has a sane default value, but it present, it should be one of the valid values.

PR #587, opened 2019-02-25
[REF] pylint-cfg: enable redefined-builtin pylint check

moylop260:
Closes https://github.com/OCA/pylint-odoo/issues/238

cc @Yajo 

FYI It was disable since that odoo uses `id` in method parameters for old api like:
 - `def copy(self, id, ...)`

Now, using new api let's to see if there are valid cases yet.
Then it new check is enabled like beta one (won't affect travis build status)


pedrobaeza:
I would put this even in 12 as required, and for 11 not so clear, but even that one is enough. If a branch goes red, it's because an invalid use. But at least for v12+.

moylop260:
It new check is not affecting travis build since that we are not sure of the result for all versions.
We can start enabling it as beta like this PR
After, we can added it like stable one for newest versions.

pedrobaeza:
What about my comment of doing it required for that versions?

Yajo:
It seems to me like others prefer the safer approach proposed by @moylop260, so we can leave it as beta for some time. This should be merged then.

pedrobaeza:
You know that most people ignore beta messages in OCA, so I still insist in doing it required for v12+

moylop260:
After work https://github.com/OCA/pylint-odoo/issues/224 we will can enable by version each check.
For now, it is not easy to do a check just for v12+

I know that we are ignoring beta messages in OCA, but it is the first valid version for now.
If you don't merge it we won't have it from beta and stable neither IMHO that is worse than just beta.

pedrobaeza:
OK, thanks for that. Merging for now in this state.

PR #594, opened 2019-04-02
[REF] travis_run_flake8.cfg: disable E203 whitespace before ':'

lmignon:
E203 is not PEP 8 compliant, we should tell Flake8 to ignore these warnings. https://www.python.org/dev/peps/pep-0008/#whitespace-in-expressions-and-statements
This change allows us to use MQT on code formatted by black and thus opens the door to the use of this tool on OCA repositories.


lmignon:
cc @simahawk @sbidoul @sebastienbeau

Issue #600, opened 2019-04-29
Odoo 12+ False success build if demo data can't install


moylop260:
After the following change from odoo for v12+: 
 - https://github.com/odoo/odoo/pull/25436

If a demo data has errors `exit` code is 0 (success) and a WARNING level is used even if there is a traceback in the log.

For runbot it is not a problem since that runbot analyze the log using:
ERROR, TRACEBACK, WARNING...
 - https://github.com/odoo/runbot/blob/4206d75256a9ca4b8d266fc56699fc2998b44ff4/runbot/models/build.py#L21-L22

For MQT is not reported since that is not analyzed the traceback and the warning are ignored.

# Steps to reproduce it:
Download the following valid odoo log output:
 - https://gist.github.com/moylop260/6aa35b6e96b2b459eb77294a9cc96dc7

Parser the log:
`cd ~/maintainer-quality-tools/travis`
`python -c "from test_server import has_test_errors;print(has_test_errors('log.txt', 'openerp_test', '12.0'))"`

Check the output:
`----------`
`0`

# Current behavior:
The count error is 0 (not detected).

# Expected behavior:
Count error should be 1 (detected) and show the traceback.

PR #601, opened 2019-04-29
[REF] has_test_errors: Detect error if there is a traceback in the log

moylop260:
Fix https://github.com/OCA/maintainer-quality-tools/issues/600


pedrobaeza:
Not sure this is a good idea, as I remember to see some tracebacks that are catched by a try and tested, and then they shouldn't go red.

moylop260:
assertRaises or/and odoo.tools.mute_logger is the solution for those kind of tests
Could you give us one real example in order to make the fix using this way?

pedrobaeza:
I see right now this one:

https://travis-ci.org/OCA/server-tools/jobs/525558942#L1635

My point is that this will turn red some branches.

PR #602, opened 2019-05-05
Pylint tests for Beta|Stable|Mature on development_status

operezavina:
### Update pylint tests
- Using the manifest file and development_status key

### Values on manifest file
- development_status:  "Alpha|Beta|Production/Stable|Mature"


max3903:
@sbidoul @pedrobaeza Any interest to get this one fixed and merged?

pedrobaeza:
I think it can be interesting, but not sure with new pre-commit in 13.0 if this can work as is. @sbidoul can tell better than me.

sbidoul:
Personally I'm not convinced lowering the quality bar for Alpha modules would facilitate contributions. Much less for Beta (which is the default level IIRC). 

The errors raised by pylint-odoo are easy to fix most of the time, and many of them are non blocking. 

IMO we don't need more low quality modules in OCA, we need more maintainers for medium and high quality modules.

So I'm rather -1 on this PR.

@pedrobaeza regarding pre-commit such thig is feasible (e.g. if we want to raise the bar for Mature modules) by creating a custom OCA pre-commit hook that selects the right .pylinrc file depending on development status.

pedrobaeza:
I was not talking about lowering quality checks indeed! I was talking about showing as errors some beta messages that now are only shown as warnings.

sbidoul:
Ah, I might have misunderstood the PR then, sorry if this is the case.

pedrobaeza:
Being honest, I'm not sure either what is the goal of the PR... Reading the code I see that possibility, but maybe it's another thing.

max3903:
This PR is mostly about separating the tests based on the 3 (or 4) maturity levels.

I am fine with increasing the tests on beta modules, as far as it is lower than stable.

PR #612, opened 2019-09-30
[REF] pylint.cfg: Disable copy-wo-api-one

moylop260:
Fix https://github.com/OCA/pylint-odoo/issues/257


pedrobaeza:
Shouldn't this be deactivated only in 13.0? Is this still relevant with the new pre-commit system?

sbidoul:
Done for 13 in 50d3a093135e379545c15a58b2a8f5e1e533e19a

PR #613, opened 2019-10-03
pre-commit support

sbidoul:
Two things in this PR:

- lint with `pre-commit` instead of the traditional flake8+pylint ~~if `.pre-commit-config.yaml` is present at repo root~~ (pre-commit is run explicitly in .travis.yml)
- allow 88 characters line length for module maintainers who'd want to use black on branches < 13

See https://odoo-community.org/groups/contributors-15/contributors-147380?mode=thread&date_begin=&date_end= for more context.


sbidoul:
Example usage in https://github.com/OCA/storage/pull/12

sbidoul:
@pedrobaeza @guewen @yvaucher @moylop260 @lmignon @sebastienbeau @gurneyalex 

In this branch I'm preparing the linting configuration file that we will push to all 13.0 branches, following our conversations during #ocadays this week.

The following files will be pushed and maintained up-to-date by the bot:
- .pre-commit-config.yaml
- .flake8
- .pylintrc
- .isort.cfg

It uses black with all default options, in order to make it as easy as possible to use it from their favorite IDE. The presence of .flake8 and .pylintrc at the repo root also makes it very easy to integrate them in IDEs.

Can you have a look, and is possible try to run it locally? It is easy.
- install pre-commit according to instruction at https://pre-commit.com/
- `git clone https://github.com/acsone/storage --branch basic-pre-commit-support`
- install the git hook script locally in your git clone with `pre-commit install`
- run it on all files with `pre-commit run -a`
- try to create a commit that breaks a lint rule and observe pre-commit gives you the error immediately

Your eagle eye on the `.pylintrc` config is most welcome. It's also the opportunity to decide if we want to enforce new checks on 13.0. @moylop260 in particular.

sbidoul:
Another example in https://github.com/OCA/queue/pull/175

pedrobaeza:
I prefer to not do quote replacing. As stated in https://github.com/OCA/partner-contact/pull/790#discussion_r331956699, it can generate big diffs and also the need to escape " characters inside the string.

It doesn't add too much value to the homogenization, as they look pretty similar.

sbidoul:
No strong opinion on quotes on my side, except it's one less thing to think about, and quote normalization is Black's default.

pedrobaeza:
Well, I always look for minimizing diff, but also I'm not very strong on this. Let's see what others say.

lmignon:
I prefer quote normalization. If we start to add exceptions, it will become an endless discussion since there will always be a good reason for some to request an exception and finally, no harmonization will be possible.
On my projects where black is now used, I find it a plus to have harmonization at this level as well. In this way, there is no more discussion on this subject and only one rule is applied by everyone, regardless of who is working on the code. Otherwise, you will very quickly find that you have lines in the same file with single quote and others with double quotes.
It is true that diff generated by the first execution of black can be scary. But for me it's an evil for the better.

pedrobaeza:
OK, seems reasonable

foutoucour:
I would like to mention that the order of import is important. I dunno if the following use case would be an issue with black:
``` python
# first import core python libraries
from datetime import datetime
from logging import getLogger

# then import third party lib
from colored_logging import getLogger

# then import local libs
from datelib import datetime
```
In the given case, and as a good practice, it is good to keep the order of section of imports. The local libs must be at the end of the imports so you are sure about the calls we do.

Is it something that the `isort` would mess with ?

PS: out of the scope of the PR, but why black and not a codacy or a codeclimate ? (sorry I didn't follow the threads on the subject)

Cheers from Canada ;)
Jordi

sbidoul:
Hi Jordi,

`isort` does change the import order yes, in a deterministic manner.
So the example you mention would be impacted if you relied on a specific import order.
I'd however argue these specific examples are bad practice and must be fixed.

The current config in this PR preserves import order in `__init__.py`, where it can be important for models registration.

If there is a real situation where import order must be preserved, `isort` has mechanism to ignore some imports or entire files.

Regarding codeclimate or codacy, at first glance they seem to be different kind of tools. Black+isort is just a code formatter, and should suffice to silence warnings provided by code format analysis tools since it is a superset of pep8.

moylop260:
> @sbidoul 
It's also the opportunity to decide if we want to enforce new checks on 13.0. @moylop260 in particular.

Could we enable lints after release this one?

What is the flow to enable new checks from configuration file?
Create new PR to MQT and waiting for the bot to propagate for all repositories?

sbidoul:
> Could we enable lints after release this one?

Yes. Of course the more code there is, the harder it is to add new lints.

> What is the flow to enable new checks from configuration file?
> Create new PR to MQT and waiting for the bot to propagate for all repositories?

Exactly. I sketched the bot algorithm there: https://github.com/OCA/oca-github-bot/issues/80

sample_files/.travis.yml
```diff
@@ -27,6 +26,42 @@ addons:
 #  - gem install bootstrap-sass
 #  - gem install compass --pre
 
+stages:
+  - linting
+  - test
+
+jobs:
```
moylop260: FYI I need improve travis2docker in order to support this new way of `.travis.yml`
Meanwhile runbot won't work (notice skipped status).

sbidoul: Ah, I hope it's not a too big change.

I had to make it this way to run pre-commit with python 3.6 while still running the Odoo tests with 3.5.

As a side benefit, we avoid running test jobs when linting fails, which saves us so Travis workers.

moylop260: Ok, the fancy travis report output is so good too.
hahaha
Let me check it...

pedrobaeza: AFAIK, now Odoo 13 main Python version is 3.6 (maybe we need to change this massively).

sbidoul: > Odoo 13 main Python version is 3.6

Are you sure? I got several echoes during OdooXP it was still 3.5.

sbidoul: > maybe we need to change this massively

In any case I was planning to reset all travis configs to default on 13.0.
It's a good occasion to realign everything and people can easily forward port per-repo specific adaptations when needed.

sbidoul: We need to be 100% sure about that. As much as I'd love to use 3.6 features, we don't want to have OCA modules don't work on some Odoo deployments.

@mart-e @odony can one of you confirm what is the minimum supported python version for Odoo 13?

sbidoul: Odoo 13' setup.py also `says >= 3.5`

Yajo: Debian 10 ships with 3.7.

This debate is not very important. I think it comes from an outdated mindset that more or less we all have inside. We have been forced to stick with python 2.7 for decades, so the versioning thing was like carved on stone in our brains.

As we make the leap for python 3.5+, it's possible, maintainable and easy to write code that works fine across 3.5, 3.6, 3.7 and 3.8. Just avoid writing `async` and `await` and you're done, basically.

So my vote goes for using the version that means less problems and maintenance. If it is 3.6, then go for it. :blush: 

<details><summary>:warning: off-topic content ahead :warning: </summary>

IMHO there's a conceptual mistake in the travis2docker thing (I already spoke about this with some of you in 2018). It's really a lot easier to do it reverse: just make travis use docker, instead of make docker imitate travis. This way, each base image for each odoo version includes the dependencies it needs, and all just works. Travis uses py3.6+ to lint, and the underlying image uses whatever python version is shipped within it. That'd be much easier.

Without trying to impose our solution, just FTR doodba has 2 different set of tags. I.e., image `tecnativa/doodba:12.0-onbuild` includes all the "doodba experience magic", but also there is `tecnativa/doodba:12.0` which only includes all dependencies needed by odoo 12.0 in a stable base. Those can be used for MQT and would make life easier for Travis and, most important, its maintainers. I'm pretty sure that all of you have similar solutions. We should flip that switch and delete travis2docker.

Back to the topic. :sweat_smile: 
</details>

sbidoul: @Yajo regarding the python version, python 3 backward compatibility can be taken for granted for sure, that is not the issue. We want OCA modules to work wherever Odoo is deployed. If Odoo says the minimum supported version is 3.5, some people somewhere will deploy it on that version and we need to test on 3.5 *to make sure OCA addons don't use 3.6+ features* that would break in some places where Odoo otherwise works fine.

sbidoul: BTW, @moylop260 whatever the answer is regarding the python version, the travis2docker update is useful, because there are Travis workload benefits in doing linting in a separate job.

sample_files/pre-commit-13.0/.pre-commit-config.yaml
```diff
@@ -0,0 +1,62 @@
...
+  - id: black
+- repo: https://github.com/pre-commit/pre-commit-hooks
+  rev: v2.3.0
+  hooks:
+  - id: trailing-whitespace
+    # exclude autogenerated files
+    exclude: /README\.rst$|\.pot?$
+  - id: end-of-file-fixer
+    # exclude autogenerated files
+    exclude: /README\.rst$|\.pot?$
```
Yajo: Why not fix EOFs in these files?

sample_files/pre-commit-13.0/.pre-commit-config.yaml
```diff
@@ -0,0 +1,62 @@
...
+  - id: debug-statements
+  - id: flake8
+    name: flake8 except __init__.py
+    exclude: /__init__\.py$
+    additional_dependencies: ["flake8-bugbear==19.8.0"]
+  - id: flake8
+    name: flake8 only __init__.py
+    args: ["--extend-ignore=F401"]  # ignore unused imports in __init__.py
+    files: /__init__\.py$
+    additional_dependencies: ["flake8-bugbear==19.8.0"]
```
Yajo: AFAIK Black enforces an opinionated version of PEP8, and flake8 checks that it's respected. However... why do we need flake8 then? Imagine a scenario where for some weird reason Black formats a file in a way that flake8 dislikes. Satisfying both linters would be confusing and time-consuming. It would be much easier to just let Black do its thing and open an issue in the Black project in case we see there's something bad.

So, can't we just drop flake8?

sample_files/pre-commit-13.0/.pre-commit-config.yaml
```diff
@@ -0,0 +1,62 @@
...
+- repo: https://github.com/pre-commit/mirrors-pylint
+  rev: v2.3.1
+  hooks:
+  - id: pylint
+    name: pylint with optional checks
+    args: ["--rcfile=.pylintrc", "--exit-zero"]
+    verbose: true
+    additional_dependencies: ["pylint-odoo==3.0.3"]
+- repo: https://github.com/pre-commit/mirrors-pylint
```
Yajo: Why is this repo duplicated? It might be OK or not, I'm just curious...

sample_files/pre-commit-13.0/.pre-commit-config.yaml
```diff
@@ -0,0 +1,62 @@
...
+  hooks:
+  - id: isort
+    name: isort except __init__.py
+    exclude: /__init__\.py$
+- repo: https://github.com/pre-commit/mirrors-eslint
+  rev: v6.5.1
+  hooks:
+  - id: eslint
+    verbose: true
+    exclude: /static/src/lib/
```
Yajo: Is this a linter or a fixer? ESLint supports both where possible, and only linter where not. That should be enabled IMHO.

Also, shouldn't this line be like this?:

```suggestion
    exclude: */static/src/lib/
```

sample_files/pre-commit-13.0/.pylintrc
```diff
@@ -0,0 +1,81 @@
...
+    file-not-used,
+    invalid-commit,
+    missing-newline-extrafiles,
+    missing-readme,
+    no-utf8-coding-comment,
+    odoo-addons-relative-import,
+    old-api7-method-defined,
+    redefined-builtin,
+    too-complex,
+    unnecessary-utf8-coding-comment
```
Yajo: Some of these are also in the mandatory file. IMHO it doesn't make sense for a check to be in both places...

sample_files/pre-commit-13.0/.eslintrc
```diff
@@ -0,0 +1,289 @@
...
+    "moment": false,
+    "odoo": false,
+    "openerp": false,
+    "self": false
+  },
+  "env": {
+    "browser": true
+  },
+  "rules": {
+    "no-alert": "warn",
```
Yajo: We should have [some `ecmaVersion` setting](https://eslint.org/docs/user-guide/configuring#specifying-parser-options) here. It's important as it would be different depending on odoo versions.

sample_files/pre-commit-13.0/.pre-commit-config.yaml
```diff
@@ -0,0 +1,62 @@
...
+- repo: https://github.com/ambv/black
+  rev: 19.3b0
+  hooks:
+  - id: black
+- repo: https://github.com/pre-commit/pre-commit-hooks
+  rev: v2.3.0
+  hooks:
+  - id: trailing-whitespace
+    # exclude autogenerated files
+    exclude: /README\.rst$|\.pot?$
```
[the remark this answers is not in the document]
sbidoul: We don't really want them, but these files are autogenerated and we don't necessarily control the generator so it's nececssary to ignore them to avoid an infinite fix/regeneration loop.

sample_files/pre-commit-13.0/.pre-commit-config.yaml
sbidoul: Same explanation as above.

sample_files/pre-commit-13.0/.pre-commit-config.yaml
sbidoul: In some cases Black cannot enforce it (eg a long string will not be split by Black). Also, flake8 does additional useful checks than pep8.

sample_files/pre-commit-13.0/.pre-commit-config.yaml
sbidoul: It's actually a regex not a glob pattern, so it's correct.

sample_files/pre-commit-13.0/.pylintrc
sbidoul: The reason they are duplicated is that `.pylintrc` (with optional and mandatory checks) can be loaded in an IDE to have it check everything (mandatory+optional), in the hope this will make optional checks more visible to contributors who otherwise never look at a green travis to see optional checks that failed.

sample_files/pre-commit-13.0/.eslintrc
```diff
@@ -0,0 +1,289 @@
...
+    "key-spacing": "off",
+    "keyword-spacing": "warn",
+    "linebreak-style": [
+      "warn",
+      "unix"
+    ],
+    "lines-around-comment": "warn",
+    "max-depth": "warn",
+    "max-len": ["warn", {
+      "code": 80,
```
Yajo: Let's match the new standard.

```suggestion
      "code": 88,
```

sample_files/pre-commit-13.0/.eslintrc
sbidoul: So far it's the same as in pylint-odoo with all errors as warnings. Feel free to propose improvements :)

sample_files/pre-commit-13.0/.eslintrc
sbidoul: 80 + bugbear B950 is the recommended black config, so it should be ok (in my experience it works nicely like this). https://github.com/psf/black#line-length

sample_files/pre-commit-13.0/.pylintrc
Yajo: OK that makes sense. A comment in the file would help clearing doubts there.

sbidoul:
@Yajo thanks for the in-depth review!

Yajo:
No, thanks to you for the in-depth work! :wink: 

There's one thing I'm missing here: a `.editorconfig` file. I used them since long ago and it integrates pretty well with almost any IDE: https://editorconfig.org/

sample_files/pre-commit-13.0/.pre-commit-config.yaml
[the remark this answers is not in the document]
Yajo: > these files are autogenerated

Then it would be great to have them autogenerated in the pre-commit loop also!

sample_files/pre-commit-13.0/.pre-commit-config.yaml
sbidoul: Which makes we think we should exclude /static/src/lib globally, right?

sample_files/pre-commit-13.0/.pre-commit-config.yaml
```diff
@@ -0,0 +1,62 @@
+exclude: "^setup/"
+default_language_version:
+  python: python3
+repos:
+- repo: https://github.com/psf/black
+  rev: 19.3b0
+  hooks:
+  - id: black
+- repo: https://github.com/pre-commit/pre-commit-hooks
```
Yajo: There are a few more hooks that seem elegible for our case. What do you think?:

1. check-case-conflict
2. check-docstring-first
3. check-executables-have-shebangs
4. check-merge-conflict
5. check-symlinks
6. check-xml
7. mixed-line-ending

sample_files/pre-commit-13.0/.pre-commit-config.yaml
guewen: If black can't fix a line because it's too long and it doesn't know where to cut it (very long word, ...), it will not change it, and as it didn't change it, it reports a success status.

sample_files/pre-commit-13.0/.pre-commit-config.yaml
guewen: The first one is used for optional checks, the second one for mandatory checks

guewen: Oh, but you mean, we could add the 2 hooks in the same `repo`. Probably?

sample_files/pre-commit-13.0/.pre-commit-config.yaml
sbidoul: Yes, makes sense. `fix-encoding-pragma --remove` sounds interesting too.

sample_files/pre-commit-13.0/.eslintrc
guewen: I confirm, 80 is the correct value since it's 80 + 10% -> 88.

Yajo: Sorry I don't understand. :confused:  This is for ESLint, not for Python, and we chose 80 to be consistent with Python, so if our new standard for Python is gonna be 88, then it should be for JS too.

sample_files/pre-commit-13.0/.pre-commit-config.yaml
sbidoul: Done. I chose `--fix=lf` for `mixed-line-ending`.

sbidoul:
Regarding `.editorconfig` can it be the subject of another PR?

Yajo:
As it's meant to be a code standarization tool that should be pushed to each repository, I think it fits fine here.

Do you want me to PR your branch to get it, so you don't have to investigate much in its syntax and features?

sample_files/pre-commit-13.0/.eslintrc
sbidoul: Ah, sorry I did not notice we where in eslint here. Fine with me then.

sbidoul:
> As it's meant to be a code standarization tool that should be pushed to each repository, I think it fits fine here.

I hope it's not controversial, because we don't want to delay this too much.

> Do you want me to PR your branch to get it, so you don't have to investigate much in its syntax and features?

Yes please.

Yajo:
Done at https://github.com/acsone/maintainer-quality-tools/pull/1.

sample_files/pre-commit-13.0/.pre-commit-config.yaml
guewen: It works: https://github.com/acsone/maintainer-quality-tools/pull/2

sample_files/.travis.yml
odony: > Are you sure? I got several echoes during OdooXP it was still 3.5.

@sbidoul Would you by any chance recall who mentioned 3.5?

We haven't changed the minimum version yet, and we globally refrained from using new features like f-strings. Yet I'm not sure we will be able to keep supporting 3.5 officially. Our 13.0 packages are all built using 3.6 or 3.7, and the runbot Dockerfile is using a single image based on Bionic so 3.6.

We've just tried, the 13.0 testsuite does not pass with 3.5 anymore. It's mostly minor things (dict key order, etc.), but it will keep breaking if we don't test with 3.5. This, plus the fact that our benchmark distros both have 3.6+ now, makes a good case for 3.6.
We'll double-check internally and @d-fence will probably update the minimum version later today or tomorrow.

sbidoul: > Would you by any chance recall who mentioned 3.5?

Not really. I think I chatted about it with @mart-e and got an indirect mention of it by a colleague. Was it @lmignon ?

@moylop260 @gurneyalex will we need to do something for the OCA runbot to use python 3.6 on the 13.0 branch?

sbidoul:
@yvaucher thanks. It's almost ready. Last things to do:
- [ ] update .travis.yml for python 3.6
- [ ] finalization of https://github.com/acsone/maintainer-quality-tools/pull/1 (@yajo)
- [ ] feedback and/or fix for travis2docker (@moylop260)

pedrobaeza:
As we are switching to 3.6, do we still need the new stages pipeline? Using previous one won't need to change travis2docker. If still needed, can we propose to use other pipeline with Docker like @Yajo said?

sample_files/.travis.yml
moylop260: @sbidoul 
We don't need to do anything extra to use python3.6 for OCA runbot.
It is pre-installed py3.5, py3.6 and py3.7 and it is used from `.travis.yml`

moylop260:
@sbidoul @gurneyalex 
I just released a new version of travis2docker in order to support jobs:

I'm testing this new way from our runbot:
 - https://runbot.vauxoo.com/runbot/repo/git-github-com-oca-storage-git-159

It is working fine!

moylop260:
Except that runbot is not supporting running docker into docker.
Do you have a project where docker is not used but the new jobs travis way is used?

sbidoul:
@moylop260 thanks! There is https://github.com/OCA/partner-contact/pull/790 but runbot still says "skipped". BTW, is there a place where some log runbot can be seen in such situation?

sample_files/.travis.yml
```diff
@@ -1,10 +1,10 @@
 language: python
-sudo: false
 cache: pip
```
yvaucher: @sbidoul 

I saw on https://github.com/OCA/queue/pull/175 that @guewen added cache for pre-commit
You also added it here https://github.com/OCA/storage/pull/12/files#diff-354f30a63fb0907d4ad57269548329e3R6

```suggestion
cache:
  directories:
  - $HOME/.cache/pip
  # Add cache for pre-commit on repo that uses it (mandatory for >= 13)
  - $HOME/.cache/pre-commit
```

Seems a good idea to cache pre-commit hooks downloads. Any reason not to include it here?

sbidoul: > Any reason not to include it here?

The only reason is that I forgot to do it :) Thanks for noticing.

moylop260:
@gurneyalex 
Could you upgrade https://pypi.org/project/travis2docker/5.1.0 for https://runbot.odoo-community.org please?

FYI I just tested the PR: https://github.com/OCA/partner-contact/pull/790 using our runbot.vauxoo.com and it is working well:
 - https://runbot.vauxoo.com/runbot/build/86193

sbidoul:
@Yajo I tweaked editorconfig a bit more, can you check? I'm not 100% comfortable pushing that because it's not been announced properly.

@Yajo what about the `ecmaScript` thingy? The current config is the one from pylint-odoo? Does it really need tweaking? If yes, can you clarify why? Also which year? 2019 sounds a bit "cutting edge".

moylop260:
@sbidoul 
FYI I just updated travis2docker to support stages like this:
 - https://github.com/OCA/storage/blob/a9b0c0f788fefd320dfb9247f8c440b1fc294542/.travis.yml

But it is not supporting the following way:
 - https://github.com/OCA/storage/blob/e9ae4bd810ec108f8a830a849ed3e58640805a01/.travis.yml#L29

Notice that there is not a `stage: test` and it is not supported yet.

sbidoul:
@moylop260 thanks. I'll push the new config with explicit jobs in the test stage.

sample_files/pre-commit-13.0/.eslintrc
sbidoul: I used 2017 as discussed in https://github.com/OCA/pylint-odoo/pull/260#discussion_r332952195

sample_files/.travis.yml
sbidoul: https://github.com/odoo/odoo/pull/38364 is approved, to 3.6 it is.

sbidoul:
Let's go! Thanks to all involved here.

pedrobaeza:
Thanks for the efforts!

What are next steps? I think we should post a blog entry with the announcement and maybe the steps to get it work for being a link for reference. Should we put this manually on each repo?

pedrobaeza:
By the way,

66479506-efc56680-ea9c-11e9-9103-82f5919e7293

:yum:

emagdalenaC2i:
Are this related with this travis red PR https://github.com/OCA/web/pull/1402 ??

sbidoul:
> What are next steps

I pushed the configs to all 13.0 branch. Only one repo needs manual intervention because of lint check that was made mandatory: https://github.com/OCA/server-env/pull/33

I updated the migration procedure to mention a new step: `pre-commit run -a` to run black and isort, then create a single commit with the code formatting.

I still need to test and document a procedure for existing 13.0 PR. It should go along these lines (if someone wants to test it first, it's welcome):

```
$ git rebase 13.0
$ pre-commit run -a
$ git add -u
$ git commit -m "black, isort"
```

I'm preparing a post.

pedrobaeza:
Runbot still appears as skipped. Isn't the @moylop260's patch applied on our runbot?

I will try the steps with an existing PR and try to prepare a blog post.

emagdalenaC2i:
Ok. I have fixed the PR OCA/web#1402 but runbot is taking a loooong time to respond

moylop260:
I will send a email to @gurneyalex in order to upgrade travis2docker on runbot

gurneyalex:
I'm on it

sbidoul:
Here is a detailed blog post about this topic: https://odoo-community.org/blog/the-oca-blog-1/post/black-isort-pre-commit-97, for reference

Yajo:
I get 403 error.

pedrobaeza:
@sbidoul I'm finishing the details about the "if you already have a migration pull request that was created before the new configuration was in place"

pedrobaeza:
Published!

PR #619, opened 2019-11-19
[IMP] clone_oca_dependencies: Clone repositories from commits SHA

ebirbe:
This improvement allows to clone repositories selecting a commit SHA.
Useful when you want to work with an specific state of the repository
that is not marked by any branch.

The syntax of the line you have to place in `oca_dependencies.txt` is as
follows:

```
server-tools git@github.com:OCA/server-tools.git 12.0 0c2c386
```

Note that you need to explicitly indicate a branch where the commit is
present because git doesn't allows to clone or fetch using directly a
commit SHA without having a configuration parameter set at server side[1].

So, when you choose to clone a repository using a commit SHA,
`clone_oca_dependencies` will fetch the commit history of the branch
you selected and then reset `HEAD` position to that commit.

[1] https://stackoverflow.com/questions/31278902/how-to-shallow-clone-a-specific-commit-with-depth-1/43136160#43136160

/cc @moylop260

### DIFF
--- a/travis/clone_oca_dependencies
+++ b/travis/clone_oca_dependencies
@@ -21,7 +21,10 @@ The expected format for oca_dependencies.txt:
 * a dependency line contains:
   - the name of the OCA project
   - (optional) the URL to the git repository (defaulting to the OCA repository)
-  - (optional) the name of the branch to use (defaulting to ${VERSION})
+  - (optional) the name of the branch to use (defaulting to ${VERSION}). It is
+    required if you want to select a commit SHA in the next parameter.
+  - (optional) the commit SHA1 to use. If you set this option you MUST specify
+    the branch
 """
 from __future__ import print_function
 import sys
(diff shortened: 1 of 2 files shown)


travis/clone_oca_dependencies
```diff
@@ -41,20 +41,20 @@ def parse_depfile(depfile, owner='OCA'):
...
             branch = os.environ.get('VERSION', '8.0')
-        if len(parts) > 1:
-            url = parts[1]
-        else:
+        if not url:
             url = 'https://github.com/%s/%s.git' % (owner, repo)
-        deps.append((repo, url, branch))
+        if commit and commit.startswith('#'):
+            # It's not a commit, it's an inline comment
+            commit = False
```
ebirbe: @moylop260 I copied this from Fentanez's version but I don't see any documentation of inline comments in oca_dependencies.txt file, what do you think about?

moylop260:
Could you update the docstring?
https://github.com/OCA/maintainer-quality-tools/blob/aaee2dca9160ffe6ba43377973e9563071134910/travis/clone_oca_dependencies#L2

And sample file, please?
https://github.com/OCA/maintainer-quality-tools/blob/aaee2dca9160ffe6ba43377973e9563071134910/sample_files/oca_dependencies.txt#L15

ebirbe:
> Could you update the docstring?
> And sample file, please?

@moylop260 changes done, could you check, please?

Coverage is giving red for no related changes.

travis/clone_oca_dependencies
moylop260: Let's to remote it since that the documentation is supporting comments at start of the line

travis/clone_oca_dependencies
```diff
@@ -41,20 +44,17 @@ def parse_depfile(depfile, owner='OCA'):
         if not line or line.startswith('#'):
             continue
         parts = line.split()
-        repo = parts[0]
-        if len(parts) > 2:
-            branch = parts[2]
-        else:
+        repo, url, branch, commit = [
+            len(parts) > i and parts[i] for i in range(4)]
```
Angelfentanez: @ebirbe Wouldn't be better to get the len of the parts outside the `for` so `len()`  is not executed in every loop?

moylop260:
@Angelfentanez 
Comment in line is not supported
One line using just comment it is supported

travis/clone_oca_dependencies
ebirbe: > Wouldn't be better to get the len of the parts outside the for so len() is not executed in every loop?

Changed, thanks.

moylop260:
@Angelfentanez 
Could you check it, please?

Somebody else likes to comment here?

Angelfentanez:
@moylop260  +1, it works with all the cases of the test file posted above except for the inline comment, but it's expected to fail in those cases since it's not supported.

Issue #626, opened 2020-01-21
Broken CI


Yajo:
This repo is broken.

The tests exhibit some kind of race condition that sometimes makes builds fail.

You can see the failures in https://travis-ci.org/OCA/maintainer-quality-tools/builds/637822768, which are unrelated to the PR.

I just rebuilt one job in the master branch and it failed with the same failure: https://travis-ci.org/OCA/maintainer-quality-tools/jobs/623695176

The failure seems to come from around the combination of these code chunks, and has been probably introduced by @hugo-ad in https://github.com/OCA/maintainer-quality-tools/pull/434:

https://github.com/OCA/maintainer-quality-tools/blob/f5651848a8670ddddfe71a29e6639de315b1b826/travis/self_tests#L255-L257

https://github.com/OCA/maintainer-quality-tools/blob/ae6dfd29dfdf00a34a38bc4c75fea0a32e1bec94/travis/test_server.py#L409-L454

I have no idea on how to fix it, this is all the info I could get :confused:

pedrobaeza:
@moylop260 you are one of the masters of this repo. Can you help?

moylop260:
I @Yajo 
Thank for your issue

FYI I see the error here:
 - https://travis-ci.org/OCA/maintainer-quality-tools/jobs/623695176#L756

I have created the following PR in order to verify it:
 - https://github.com/OCA/maintainer-quality-tools/pull/627

Yajo:
Just in case it helps, let me explain how I got to the conclusion I said in https://github.com/OCA/maintainer-quality-tools/issues/626#issue-552863044.

AFAICS, comparing the :x: job with the latest :heavy_check_mark: one, the main difference is here:

[Job that failed](https://travis-ci.org/OCA/maintainer-quality-tools/jobs/623695176#L722-L726):

```
test_modules_to_preinstall (__main__.MainTest) ... ok
test_pylint_check (__main__.MainTest)
Testing empty paths and pylint_run fix of: ... skipped 'Set LINT_CHECK'
test_travis_helpers (__main__.MainTest) ... ----------
ok
```

[Job that succeeded](https://travis-ci.org/OCA/maintainer-quality-tools/jobs/614129658):

```
test_modules_to_preinstall (__main__.MainTest) ... ok
----------
test_pylint_check (__main__.MainTest)
Testing empty paths and pylint_run fix of: ... skipped 'Set LINT_CHECK'
[...]

PR #627, opened 2020-01-21
[REF] install: pin coveralls version and supporting broken_uninstallable module

moylop260:
Related with https://github.com/OCA/maintainer-quality-tools/issues/626
cc @Yajo 

And Supporting broken_uninstallable module
After new logger.error from https://github.com/odoo/odoo/commit/c5d6a39

### DIFF
--- a/sample_files/.gitlab-ci-2.yml
+++ b/sample_files/.gitlab-ci-2.yml
@@ -22,7 +22,7 @@ variables:
 before_script:
   - curl -sL https://deb.nodesource.com/setup_10.x | bash -
   - apt-get install -y  sudo postgresql-client expect-dev python-lxml nodejs python-dev python-pip build-essential libsasl2-dev python-dev libldap2-dev libssl-dev
-  - pip install coverage coveralls codecov
+  - pip install codecov==2.0.15 coveralls==1.8.2 coverage
 
 lint:
   stage: test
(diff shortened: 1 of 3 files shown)


Yajo:
So, do you mean that the culprit is the version of coveralls?

sample_files/.gitlab-ci-2.yml
```diff
@@ -22,7 +22,6 @@ variables:
 before_script:
   - curl -sL https://deb.nodesource.com/setup_10.x | bash -
   - apt-get install -y  sudo postgresql-client expect-dev python-lxml nodejs python-dev python-pip build-essential libsasl2-dev python-dev libldap2-dev libssl-dev
-  - pip install coverage coveralls codecov
```
pedrobaeza: Shouldn't this install also pinned versions?

moylop260:
> @Yajo 
So, do you mean that the culprit is the version of coveralls?

I saw something related and I just tested pin it

But I didn't looking for more details.

Maybe a recent changes incompatible.

pedrobaeza:
Wasn't this green previously?

moylop260:
Yes, the coverage error was fixed but at the same time a new change was introduced in Odoo to log an new error:
 - https://github.com/odoo/odoo/commit/c5d6a3977de85fb974a4940fb11d54ef847e08e4#diff-2f47c68d06883cfb9f76e474abca7676

So, the new red is unrelated.

moylop260:
Now it is green... until now 😄

pedrobaeza:
Hehe, let's merge it before turning again :yum:

PR #628, opened 2020-01-24
[FIX] editorconfig syntax fix

Yajo:
A couple of fixes missing from #618.

@Tecnativa TT20357


pedrobaeza:
Can you explain what these things do?

Yajo:
Yes, sorry. Syntax was wrong and it wasn't doing what we expected. The comma has no special meaning outside curly braces.

OTOH there's no `.eslintrc` file anymore.

pedrobaeza:
OK, fast-tracking then.

PR #629, opened 2020-01-27
Add isort third party detection from requirements.txt

### LINKED ISSUE #625: .isort.cfg is a conflicts bottleneck
In 13.0 projects, when we run pre-commit and an addon uses a new python lib, `seed-isort-config` will add the lib in the file `.isort.cfg` to list it as "known third party". Example of resulting file:

```
[settings]
; see https://github.com/psf/black
multi_line_output=3
include_trailing_comma=True
force_grid_wrap=0
combine_as_imports=True
use_parentheses=True
line_length=88
known_odoo=odoo
known_odoo_addons=odoo.addons
sections=FUTURE,STDLIB,THIRDPARTY,ODOO,ODOO_ADDONS,FIRSTPARTY,LOCALFOLDER
known_third_party=setuptools
```

This file is a huge contender for conflicts when working with unmerged pull requests and, for instance, we use gitaggregator to merge them together.

A real life example: PR https://github.com/OCA/stock-logistics-warehouse/pull/826 adds the `cubiscan` and `mock` libs, PR https://github.com/OCA/stock-logistics-warehouse/pull/797 adds `aiohttp`. On aggregation, we'll have a conflict such as:

```
diff --cc .isort.cfg
index 969be420,6bcf3847..00000000
--- a/.isort.cfg
+++ b/.isort.cfg
@@@ -9,4 -9,4 +9,10 @@@ line_length=8
  known_odoo=odoo
  known_odoo_addons=odoo.addons
  sections=FUTURE,STDLIB,THIRDPARTY,ODOO,ODOO_ADDONS,FIRSTPARTY,LOCALFOLDER
++<<<<<<< HEAD
 +known_third_party=aiohttp,setuptools
++||||||| 5ded73cf
++known_third_party=setuptools
++=======
+ known_third_party=cubiscan,mock,setuptools
++>>>>>>> e7f58dcdc68ccd1d53389b1fe03ec94a796f830c
```

Of course, conflicts between pull requests are not new and are inevitable, but we generally try to avoid them. The situation with `.isort.cfg` is painful though.

Is ordering the imports so important? (I find the whole `isort` "infrastructure" complex for the added value)

Thoughts?

guewen:
The current setup used in the OCA for isort is done by using 2
pre-commit hooks:

1. seed-isort-config: find third-party libs and updates the
'.isort.cfg' file's "known_third_party" option with the list
2. isort: runs isort, which will use the list of third party
libs provided by 'seed-isort-config' in the "known_third_party" option.

This is an issue [0] with the workflow adopted by many contributors of
the OCA: as the pull requests may need some time to be merged, we use
temporary branches where we aggregate the various PRs.

If several pull requests add a third party library, we'll have a
conflict, and when the process of merging these pull requests is
automated (e.g. with git-aggregator [1]), it becomes tedious.

The list of libs in the "known_third_party" variable is not even
editable manually, as "seed-isort-config" overwrite the value on every
commit: it doesn't really make sense to actually store this.

As pointed out by @sbidoul [2], isort mentions in their changelog:

> Support for using requirements files to auto determine third-paty
> section if pipreqs & requirementslib are installed.

Which is the goal of this change, details:

* Remove 'seed-isort-config' from pre-commit as the list of libs will be
  provided by requirements.txt
* Replace the isort pre-commit repo by the official repo (the mirror says
  it is been deprecated in favor of the official repo)
* Adds pipreqs and pip-api in additional dependencies to activate
  isort's feature to read requirements.txt [3]
* Adds types: [python] in the pre-commit config: the mirror had it and
  the official repo doesn't. Without it, some files such as .pot are
  modified (different EOF). (it can be removed after the next isort
  release)
* Add an union merge driver on 'requirements.txt' to resolve conflicts
  on this file (on conflicts, both lines are kept, which works in most
  cases on this file, in the rare case it's an issue because we have
  2 requirements for different versions, we'll have to fix manually)

Note: it is now important to have the Python libraries declared in
"requirements.txt" to have them ordered in the proper place.

Alternatively, we could have tried 'reorder-python-imports' in place of 'isort' [4].

This setup has been tested successfully on https://github.com/OCA/stock-logistics-warehouse/pull/828

[0] OCA/maintainer-quality-tools#625
[1] https://github.com/acsone/git-aggregator
[2] OCA/maintainer-quality-tools#625 (comment)
[3] https://github.com/timothycrosley/isort/blob/500bafabbd51a6005c11a00c4738a2438990e48a/pyproject.toml#L42
[4] https://github.com/asottile/reorder_python_imports

Closes #625

### DIFF
--- a/sample_files/pre-commit-13.0/.isort.cfg
+++ b/sample_files/pre-commit-13.0/.isort.cfg
@@ -9,4 +9,3 @@ line_length=88
 known_odoo=odoo
 known_odoo_addons=odoo.addons
 sections=FUTURE,STDLIB,THIRDPARTY,ODOO,ODOO_ADDONS,FIRSTPARTY,LOCALFOLDER
-known_third_party=
--- a/sample_files/pre-commit-13.0/.pre-commit-config.yaml
+++ b/sample_files/pre-commit-13.0/.pre-commit-config.yaml
@@ -85,13 +85,12 @@ repos:
     rev: v1.26.2
     hooks:
       - id: pyupgrade
-  - repo: https://github.com/asottile/seed-isort-config
-    rev: v1.9.4
-    hooks:
-      - id: seed-isort-config
-  - repo: https://github.com/pre-commit/mirrors-isort
-    rev: v4.3.21
+  - repo: https://github.com/timothycrosley/isort
+    rev: 4.3.21-2
     hooks:
       - id: isort
         name: isort except __init__.py
         exclude: /__init__\.py$
+        # TODO, the next time we upgrade isort's rev, we should remove 'types'
+        types: [python]
+        additional_dependencies: ["pipreqs==0.4.10", "pip-api==0.0.13"]
--- a/sample_files/pre-commit-13.0/.gitattributes
+++ b/sample_files/pre-commit-13.0/.gitattributes
@@ -1,0 +1,2 @@
+# Limit conflicts in PRs by automatically merging lines of both sides
+requirements.txt merge=union


sample_files/pre-commit-13.0/.pre-commit-config.yaml
```diff
@@ -85,13 +85,11 @@ repos:
...
-      - id: seed-isort-config
-  - repo: https://github.com/pre-commit/mirrors-isort
-    rev: v4.3.21
+  - repo: https://github.com/timothycrosley/isort
+    rev: 4.3.21-2
     hooks:
       - id: isort
         name: isort except __init__.py
         exclude: /__init__\.py$
+        'types': [python]
```
Yajo:
```suggestion
        # TODO Remove types after upgrading isort
        types: [python]
```

sample_files/pre-commit-13.0/.pre-commit-config.yaml
```diff
@@ -85,13 +85,11 @@ repos:
...
+        additional_dependencies: [pipreqs, pip-api]
```
Yajo: Pin these versions please.

sample_files/pre-commit-13.0/requirements.txt
```diff
@@ -0,0 +1,2 @@
+# Add the python requirements here.
+# Note: the libs in this file will be considered as "third-party libs" by isort.
```
Yajo: Not sure if this diff is good. It's going to introduce many conflicts probably, and as long as automations are transparent to the user, like the one you're proposing here, further explanations don't help much.

If you're gonna introduce instructions here, then IMHO it's better to just put a link, like "See http://whatever", leading to the corresponding section in our contributors guideline, where full explanations can be written.

sample_files/pre-commit-13.0/.pre-commit-config.yaml
```diff
@@ -85,13 +85,11 @@ repos:
...
-    rev: v1.9.4
-    hooks:
-      - id: seed-isort-config
-  - repo: https://github.com/pre-commit/mirrors-isort
-    rev: v4.3.21
+  - repo: https://github.com/timothycrosley/isort
+    rev: 4.3.21-2
     hooks:
       - id: isort
         name: isort except __init__.py
```
Yajo: I wonder... Why do we ignore these files with isort? Don't we want sorted imports there too? :thinking: 

cc @sbidoul

sbidoul: I think I made it like this because there are situations where the import order of models classes is important for the construction of the Odoo registry.

sample_files/pre-commit-13.0/requirements.txt
guewen: So should I only put an empty `requirements.txt` file?

guewen:
@Yajo thanks for your review, I pushed a fixup with the updates.

sample_files/pre-commit-13.0/.pre-commit-config.yaml
Yajo: I guess that was only back then in v7, where one had to instantiate models after declaration.

If there are isort exclusion tags, I'd vote for sorting these too. In corner cases, we just add those exclusion tags and go on.

sample_files/pre-commit-13.0/.pre-commit-config.yaml
```diff
@@ -85,13 +85,12 @@ repos:
...
         exclude: /__init__\.py$
+        # TODO, the next time we upgrade isort's rev, we should remove 'types'
+        types: [python]
+        additional_dependencies: ["pipreqs==0.4.10", "pip-api==0.0.13"]
```
Yajo: Don't you need to add requirements?

See https://github.com/timothycrosley/isort#installing-isort

sample_files/pre-commit-13.0/.pre-commit-config.yaml
pedrobaeza: I remember at least one case where loading order of models matters (I think it's because defaults applied on fields), so I prefer to not automate this.

sample_files/pre-commit-13.0/.pre-commit-config.yaml
guewen: `requirements` is a sort of alias on `pipreqs` + `pip-api`: https://github.com/timothycrosley/isort/blob/0882c8885b88618ea55b97ace256cdf833a1547d/pyproject.toml#L42

guewen:
Squashed

sbidoul:
@guewen One issue I detected with this is that the usual Odoo dependencies (e.g. lxml, dateutil) are usually not declared in `requirements.txt` and are now moved to the local imports group by isort.

Ah well... nothing is simple.

Yajo:
Do you have an example?

sbidoul:
For example, any addon that imports `dateutil` will have it misplaced by isort, if `known_thirparty` is not set in `.isort.cfg`.

Putting dateutil in the manifest external deps does not help, because what lands in `requirements.txt` is `python-dateutil` (which is correct) and not `dateutil`, and `isort` therefore does not detect it as third party.

One way to mitigate this is to have the list of common dependencies in `known_thirdparty` (it could be the list that is also known by pylint-odoo).

That would not automate it fully, as when a third party package has an import name different than it's distribution name, it would need to be added manually to `known_thirdparty`.

I'm experimenting with that in https://github.com/OCA/mis-builder/pull/260.

BTW, I also wrote a small tool to generate `requirements.txt` from the manifests in case that ends up being useful: https://github.com/acsone/setuptools-odoo/pull/42

Yajo:
OK I have another idea.

If we know the `ODOO` and `ODOO_ADDONS` sections... then can't we assume all other modules are `THIRDPARTY`?

[isort docs](https://github.com/timothycrosley/isort/wiki/isort-Settings#full-reference-of-isort-settings) say:

> **default_section**: The default section to place imports in, if their section can not be automatically determined. `FIRSTPARTY`, `THIRDPARTY`, etc.

So if we add to `.isort.cfg`:

```cfg
default_section=THIRDPARTY
```

We should be able to use isort exactly the same way without needing to seed it. No conflicts, same behavior. Could this be our missing silver bullet?

Yajo:
My hypothesis is confirmed, that's the path to go.

See a real world scenario: https://github.com/OCA/product-attribute/pull/582

sbidoul:
Yes! Nice find @Yajo

pedrobaeza:
So are we deploying this config in all repos together with the prettier-xml thing?

sbidoul:
> So are we deploying this config in all repos together with the prettier-xml thing?

@pedrobaeza That is what I intend to do yes, when #632 and #636 are cleared.

pedrobaeza:
OK, thanks. @Yajo can you help in finishing both?

sample_files/pre-commit-13.0/.pre-commit-config.yaml
Yajo: This is possibly not needed now.

sample_files/pre-commit-13.0/.isort.cfg
```diff
@@ -9,4 +9,3 @@ line_length=88
 known_odoo=odoo
 known_odoo_addons=odoo.addons
 sections=FUTURE,STDLIB,THIRDPARTY,ODOO,ODOO_ADDONS,FIRSTPARTY,LOCALFOLDER
-known_third_party=
```
Yajo: Missing the default section, see https://github.com/OCA/maintainer-quality-tools/pull/629#issuecomment-588131706.

sbidoul:
@Yajo this one is superseded by #639, if @guewen agrees.

guewen:
Yes, thanks folks!

kittiu:
Hi, by looking around seem that now we use THIRDPARTY in stead of define own known_thirdparty, right?

We got travis error in this PR https://github.com/OCA/server-tools/pull/1684 as it need `openpyxl`

Where can I add it? Thanks!

sbidoul:
@kittiu the issue you are having in that PR is not related to this. `.isort.cfg` is only used to control sorting of imports in python files, not for installation. I provided a solution as a comment in your the PR.

PR #630, opened 2020-01-27
Update eslintrc configuration

Yajo:
- Sort file alphabetically to make maintenance easier
- Remove more rules disabled in https://github.com/prettier/eslint-config-prettier/blob/v6.9.0/index.js
- Remove `false` and put `readonly` in globals; [`false` is deprecated](https://eslint.org/docs/user-guide/configuring#specifying-globals)
- All failures were tagged as warn because until today, pylint-odoo only printed warnings in JS files. IMHO it's time to flip the switch, so I manually enabled some checks as errors.

@Tecnativa TT20357

### DIFF
--- a/sample_files/pre-commit-13.0/.eslintrc.yml
+++ b/sample_files/pre-commit-13.0/.eslintrc.yml
@@ -5,56 +5,74 @@ env:
 parserOptions:
   ecmaVersion: 2017
 
-# Globals available in Odoo that shouldn't produce warnings
+# Globals available in Odoo that shouldn't produce errorings
 globals:
-  $: false
-  _: false
-  fuzzy: false
-  jQuery: false
-  moment: false
-  Promise: false
-  odoo: false
-  openerp: false
+  _: readonly
+  $: readonly
+  fuzzy: readonly
+  jQuery: readonly
+  moment: readonly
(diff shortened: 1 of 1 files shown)


Yajo:
Could we merge please to unblock https://github.com/OCA/web/pull/1499 and https://github.com/OCA/website/pull/689?

Tardo:
I will review this PR, wait for me before merge it please. Thanks.

sample_files/pre-commit-13.0/.eslintrc.yml
```diff
@@ -5,56 +5,74 @@ env:
...
+  array-callback-return: warn
+  callback-return: warn
+  capitalized-comments:
+    - warn
+    - always
+    - ignoreConsecutiveComments: true
+      ignoreInlineComments: true
+  complexity:
+    - warn
+    - 15
```
Tardo: Perhaps 11 is a better value? See: https://github.com/eslintcc/eslintcc#complexity-ranks

Yajo: Humm I guess we'll have to see this in practice to judge, just by seeing the numbers I'm not very able to guess it.

PR #631, opened 2020-01-30
[IMP] Use verbose colored pre-commit in CI

Yajo:
After https://github.com/OCA/maintainer-quality-tools/pull/630, some linters like eslint may produce warnings that would be unnoticed if no errors were found.

To be able to still see the warnings, I add `--verbose` to pre-commit.

Also, I add the handful `--color always` to make colored output even without a TTy, as it was the case.

@Tecnativa TT20357

### DIFF
--- a/sample_files/.travis.yml
+++ b/sample_files/.travis.yml
@@ -42,7 +42,7 @@ jobs:
       name: "pre-commit"
       python: "3.6"  # Black requires python >= 3.6
       install: pip install pre-commit
-      script: pre-commit run --all --show-diff-on-failure
+      script: pre-commit run --all --show-diff-on-failure --verbose --color always
       after_success:
       before_install:
     # use this linting job for repos that do not use pre-commit (default for < 13)


sbidoul:
> Not sure if this is going to be applied for all the branches. @sbidoul?

There is no magical tool that does that :)

pedrobaeza:
But you had a way to deploy to all branches all the pre-commit thing. Why it can't be used in this case?

sbidoul:
The first time was easy, we had almost empty 13.0 branches and all I had to do was to add and replace files. Now, there are local adaptations (in .travis.yml, notably) and we can't simply overwrite everything.

pedrobaeza:
Well, I do the proper replacement for each new branch, changing local things (for example, 12.0 by 13.0) in https://github.com/OCA/maintainer-tools/blob/master/tools/migrate_branch_empty.py.

If your tool is not able to, I can program it (but this is specifically each time). The question is: should we do it massively?

Yajo:
> There is no magical tool that does that :)

I guess there should be one... :thinking: 

We do have a ton of repos, each one with lots of branches... There should be a way to automate all this...

Well, it's another issue in any case; I guess a sed command would do the trick for the case at hand if necessary.

pedrobaeza:
@sbidoul is it OK for you then this change? I will handle the massive change in 13.0 branches

sbidoul:
It's annoying that this repo is still red.

pedrobaeza:
The massive changer is in

https://github.com/OCA/maintainer-tools/pull/442

and I'm applying it to all 13.0 branches with the command:
```bash
oca-massive-change 13.0 .travis.yml "pre-commit run --all --show-diff-on-failure\n" "pre-commit run --all --show-diff-on-failure --verbose --color always\n" -e oca-git-bot@odoo-community.org -m "[UPD] .travis.yml: Use verbose colored pre-commit [skip ci]"
```

Everything is performed in GitHub, with no need to clone branches.

PR #641, opened 2020-03-03
Ignore `lib/` wherever it stays

simahawk:
Libs can be place in several paths depending on project's needs.
We should always ignore them.

### DIFF
--- a/sample_files/pre-commit-13.0/.pre-commit-config.yaml
+++ b/sample_files/pre-commit-13.0/.pre-commit-config.yaml
@@ -5,7 +5,7 @@ exclude: |
   # Maybe reactivate this when all README files include prettier ignore tags?
   ^README\.md$|
   # Library files can have extraneous formatting (even minimized)
-  /static/(src/)?lib/|
+  lib/|
   # Repos using Sphinx to generate docs don't need prettying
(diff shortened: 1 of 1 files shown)


sbidoul:
@simahawk do you have a concrete example?

simahawk:
@sbidoul because you might want to put your static code inside something more meaningful instead of `src` which is there only because odoo assumes you use it when it looks up for JS translations.
Our use case is the WMS app: we serve `static/wms`  and we like to sort of  namespace it's path.

Not really mandatory but nice to have.
Also, I don't see which issue could you get by ignore `lib` wherever it stays. I think is the contrary.

sample_files/pre-commit-13.0/.pre-commit-config.yaml
```diff
@@ -5,7 +5,7 @@ exclude: |
   # Maybe reactivate this when all README files include prettier ignore tags?
   ^README\.md$|
   # Library files can have extraneous formatting (even minimized)
-  /static/(src/)?lib/|
+  lib/|
```
dreispt, line 8: Can we still enforce the `static/` prefix?
Something like `static/.*/lib/`

PR #643, opened 2020-03-11
sample files: generate setup.py files in pre-commit

sbidoul:
This will ensure setup.py files are generated on pull requests and avoid one bot commit.


moylop260:
FYI It will produce x2 the code diff for each MR
I'm not sure if it is a correct way to compile the `setup.py` for each commit.

But let's to see what about.

sbidoul:
@moylop260 what do you mean?

moylop260:
1. Create a change in `module_x/models/model_a.py`
2. Run `git commit `
3.  pre-commit creates changes in ...

Forget it, meanwhile I wrote this comment I just realized that it is using a symbolic link.

PR #646, opened 2020-03-12
Add opt-in mechanism to install dependencies with pip

sbidoul:
Setting a `MQT_DEP=PIP` environment variable in `.travis.yml` instructs MQT to discover and install addon dependencies using pip instead of `oca_dependencies.txt` and `requirements.txt`.

Specific test requirements such as specific branches of dependencies can be installed by creating a `test-requirements.txt` file at the repo root.

This is a new, opt-in, version of #343.

Example: https://travis-ci.org/github/acsone/mis-builder/jobs/661629796

### DIFF
--- a/travis/travis_install_nightly
+++ b/travis/travis_install_nightly
@@ -113,13 +113,22 @@ pip install -q QUnitSuite codecov==2.0.15 coveralls==1.8.2
 # Use reference .coveragerc
 cp ${HOME}/maintainer-quality-tools/cfg/.coveragerc .
 
-echo "Getting addons dependencies"
-clone_oca_dependencies
-clone_result=$?
-if [ "$clone_result" != "0"  ]; then
-    echo "Error cloning dependencies"
-    exit $clone_result
-fi;
+MQT_DEP=${MQT_DEP:-OCA}
+if [[ "${MQT_DEP}" == "OCA" ]] ; then
+    echo "Getting addons dependencies"
+    clone_oca_dependencies
+else
+    echo "Installing addons to test and their dependencies"
+    if [[ "${VERSION}" == "10.0" || "${VERSION}" == "9.0" || "${VERSION}" == "8.0" ]] ; then
+        pip install odoo-autodiscover
+    fi
+    for addon in $(ls setup/ -I README -I _metapackage) ; do
+        echo "-e ./setup/${addon}" >> test-requirements.txt
+    done
+    # use OCA wheelhouse because it is slightly fresher than PyPI
+    pip install --pre -e ${ODOO_PATH} -r test-requirements.txt \
+        --extra-index-url https://wheelhouse.odoo-community.org/oca-simple/
+fi
 
 if [[ "${WKHTMLTOPDF_VERSION}" == "" && $(which wkhtmltopdf) != "" ]]; then
     echo "You have installed wkhtmltopdf but is not patched (probably) then we will overwrite it"


sbidoul:
FTR, what prompted me to work on this again is that mis-builder 10.0 and 11.0 branches are red again due to broken (yet not used) dependencies in other repos. Using this branch fixes that.

sbidoul:
The reason it fixes such issues is that only the strictly necessary dependencies are installed instead of all addons of dependent repos.

travis/travis_install_nightly
```diff
@@ -113,13 +113,22 @@ pip install -q QUnitSuite codecov==2.0.15 coveralls==1.8.2
...
-clone_oca_dependencies
-clone_result=$?
-if [ "$clone_result" != "0"  ]; then
-    echo "Error cloning dependencies"
-    exit $clone_result
-fi;
+MQT_DEP=${MQT_DEP:-OCA}
+if [[ "${MQT_DEP}" == "OCA" ]] ; then
+    echo "Getting addons dependencies"
+    clone_oca_dependencies
```
sbidoul: I removed the exit code test because this script runs with `set -e` so it is dead code.

sbidoul:
Since this code is fully opt-in I plan to merge this soon-ish to resolve repair broken builds.

Unless there are strong objections of course.

simahawk:
very nice, thanks!

simahawk:
@sbidoul if we have to include pending merges shall we use `-e` + git url into test-requirements.txt then?

sbidoul:
@simahawk yes exactly. 

Note `-e` is not necessary (`travis_install_nightly` uses `-e` so `coverage` finds the source code). 

So the modern (PEP 508 / PEP 440) syntax for such a line in `test-requirements.txt` would be `odoo{version}-addon-{module_name} @ git+https://github.com/OCA/{repo}@refs/pull/{NNN}/head#subdirectory=setup/{module_name}`.

Writing this it occurs to me there could be an issue if the pending merge is for an existing addon in the same repo, as we would end up with double requirements. That situation could be detected https://github.com/OCA/maintainer-quality-tools/blob/ca231f41c81462b8916134f33c9f0c250d67457d/travis/travis_install_nightly#L120-L123.

simahawk:
> Writing this it occurs to me there could be an issue if the pending merge is for an existing addon in the same repo, as we would end up with double requirements.

which means that ATM this is not supported, correct?

I wonder: how the addons path is handled? Because if you get the python package path in the right order for `--addons-path` then this is a non-issue. No?

sbidoul:
> which means that ATM this is not supported, correct?

ATM it works only when you add test-requirements from other repos.

> how the addons path is handled

Addons path is automatic. Odoo [looks at all the parts of the `odoo.addons` namespace package](https://github.com/odoo/odoo/blob/d93ccdc5a25c061468870359ad8f749496140d19/odoo/modules/module.py#L109), which is what is populated by pip install (via setuptools-odoo).

> Because if you get the python package path in the right order for --addons-path then this is a non-issue. No?

Actually pip complains if it receives the same requirement more than once in the same `pip install` command. Even if it would not I'd say relying on addons-path order is fragile.

simahawk:
I see, thanks for the clarification! :raised_hands: 
Last question: is this documented somewhere? :innocent:

sbidoul:
> is this documented somewhere?

Depends what you mean by _this_. The best we have is the README of https://pypi.org/project/setuptools-odoo/. Now that I re-read it I notice it would benefit of a section on the principle of operation. And perhaps moving the parts that are specific to odoo<10 out of the way.

Regarding `MQT_DEP=PIP` and `tests-requirements.txt` I'm planning to post something about it soonish. Perhaps we can experiment with it on some repos, and then make it the default for v14.

PR #647, opened 2020-03-14
[FIX] do not set addons_path when installing with pip

sbidoul:
We don't need an `addons_path` when addons are installed with pip,
and if we have the same addon in several places in the `addons_path`, this confuse
.pot file generation.

Follow-up from #646


sbidoul:
Ok, this [works](https://github.com/OCA/mis-builder/commit/c5462d93b846b7e1ede305fa8be1cc6f9158f46f).

PR #650, opened 2020-03-20
do not install odoo-autodiscover in MQT_DEP=PIP mode

sbidoul:
It confuses runbot (or is it travis2docker?).

It is not needed in this specific case because odoo is installed
with -e too and comes first in sys.path so it's `__init__.py` is imported first.


sbidoul:
fast tracking as still experimental and hidden behind a feature flag

PR #653, opened 2020-03-22
remove dependency on configparser

sbidoul:
use py2/py3 compatible code

`configparser` is the only dependency that was needed in addition to regular Odoo dependencies in order to run `test_server.py`. Removing it makes it possible to avoid MQT's `requirements.txt` in most cases.


moylop260:
Merging since that the 8 builds running `TESTS=1` using python2.7, python3.5 and python3.7 is working well in the unittests

 - imagen

 - imagen

Thanks!

PR #654, opened 2020-03-23
install less dependencies in MQT_DEP=PIP mode

sbidoul:
- Move some installations steps into MQT_DEP=OCA code path (no behaviour change in that mode).
- Do not install mqt's `requirements.txt` in MQT_DEP=PIP mode.
- Generate potentially missing `setup.py` files in MQT_DEP=PIP mode.


sbidoul:
Ok, this [works](https://travis-ci.com/github/acsone/l10n-italy/jobs/301301151).

sbidoul:
Merging this: the use code path is unchanged. Only MQT_DEP=PIP mode changes and installs less normally unused dependencies.

PR #655, opened 2020-03-29
sample files: upgrade prettier

sbidoul:
Now that the prettier pre-commit hook properly supports plugins, let's use it.

### DIFF
--- a/sample_files/pre-commit-13.0/.pre-commit-config.yaml
+++ b/sample_files/pre-commit-13.0/.pre-commit-config.yaml
@@ -18,20 +18,26 @@ repos:
     hooks:
       - id: black
   - repo: https://github.com/prettier/prettier
-    rev: "1.19.1"
+    rev: "2.0.2"
     hooks:
       - id: prettier
-  # TODO Avoid awebdeveloper/pre-commit-prettier if possible
-  # HACK https://github.com/prettier/prettier/issues/7407
-  - repo: https://github.com/awebdeveloper/pre-commit-prettier
-    rev: v0.0.1
-    hooks:
-      - id: prettier
-        name: prettier xml plugin
         additional_dependencies:
-          - "prettier@1.19.1"
           - "@prettier/plugin-xml@0.7.2"
-        files: \.xml$
+        args:
+          - --plugin=@prettier/plugin-xml
+        files: "\\.(\
+          css|less|scss\
+          |graphql|gql\
+          |html\
+          |js|jsx\
+          |json\
+          |md|markdown|mdown|mkdn\
+          |mdx\
+          |ts|tsx\
+          |vue\
+          |yaml|yml\
+          |xml\
+          )$"
   - repo: https://github.com/pre-commit/mirrors-eslint
     rev: v6.8.0
     hooks:


sbidoul:
@Yajo unfortunately, when we apply this it reformats js files to add a whitespace in `functions ()`. Any idea why?

Yajo:
Yes, it's an upgrade: https://github.com/prettier/prettier/pull/3903
I found it weird without that while reviewing some PRs. I guess this kind of diff will be common when we upgrade our formatters, but seems good to me. The patch will remove [some distracting warnings](https://travis-ci.org/github/OCA/web/jobs/666748366#L273) we had due to having this step split in 2.

sample_files/pre-commit-13.0/.pre-commit-config.yaml
```diff
@@ -18,20 +18,26 @@ repos:
...
+          |html\
+          |js|jsx\
+          |json\
+          |md|markdown|mdown|mkdn\
+          |mdx\
+          |ts|tsx\
+          |vue\
+          |yaml|yml\
+          |xml\
+          )$"
```
Yajo: This is the only part I'd like to avoid. We can do it like this:

```yaml
   - repo: https://github.com/prettier/prettier

    rev: "2.0.2"
    hooks:
      - &prettier
        id: prettier
         additional_dependencies:
           - "@prettier/plugin-xml@0.7.2"
        args:
          - --plugin=@prettier/plugin-xml
      - <<: *prettier
        files: \\.xml$
```

Basically we'd be launching the same hook twice: once for the standard files supported by prettier, and once just for XML files.

sbidoul: The problem with the prettier run without the xml plugin is that it warns about an unknown option: `[warn] Ignored unknown option { xmlWhitespaceSensitivity: "ignore" }. Did you mean htmlWhitespaceSensitivity?`.

Also I find it good to be explicit about the file types we put in prettier's hands.

Yajo: > The problem with the prettier run without the xml plugin is that it warns about an unknown option: `[warn] Ignored unknown option { xmlWhitespaceSensitivity: "ignore" }. Did you mean htmlWhitespaceSensitivity?`.

It wouldn't warn because the plugin would be loaded in both steps. Just no XML files would be selected in the 1st one.

> Also I find it good to be explicit about the file types we put in prettier's hands.

We don't do it for any other formatter, unless there's a problem with Prettier's selection that can't be fixed with exclude, I prefer to keep this shorter and more maintainable, just like it is.

sbidoul:
The js formating change is very annoying to me. We must strive for stability here because each update has signigicant impact on contributors (merge confliicts, IDE configuration, forward/backport issues etc).

Can we configure that specific whitespace?

Yajo:
Nope: https://github.com/prettier/prettier/pull/3903#issuecomment-602754118

You're updating to a new major version and that's expected. It would happen the same with black for example. Otherwise just prepare this change for Odoo 14 and leave 13 branch as is. :man_shrugging:

sbidoul:
Ok, closing for now, then.

pedrobaeza:
Then should we maintain always a previous "prettified" commit on each migration?

By the way, I have added prettier to the migration instructions:

https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-13.0/_compare/564c9e48e86a1470541375662f45b4adab986c6a...1ea87d50578a44569eee600a2b7f66936f3514dd

Yajo:
Good, IMHO that's all.

We should upgrade formatters only when there's a new odoo release then.

PR #656, opened 2020-04-01
[FIX] Add pylint_odoo dependency to requirements.txt

dreispt:
If you `pip install -r requirements.txt` and then try to run `git/pre-commit`, it will error because pylint is missing.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,5 @@
 simplejson
+pylint_odoo
 pyserial
 pyopenssl>=16.2.0
 pyyaml==3.12 ; python_version < '3.7'


sbidoul:
HI @dreispt this should not be necessary for pre-commit to run correctly, since pre-commit creates an isolated environment for each hook it runs.

Can you provide more details about your use case and problem?

dreispt:
There is a `git/pre-commit` script. I'm just running it, at the lack of better guidance.

pedrobaeza:
That one is very old and previous to all the new precommit stuff for v13 (same name, different tool). Any way, it contains instructions for making it run in that old environments that includes the installation of pylint.

sbidoul:
Ah, I understand. Can't we just remove that git directory and point people to the sample_files/pre-commit directory?

dreispt:
Thank you Pedro. I looked again and it is the only pre-commit file I can find. Can you point at the newer one you refer?

I also can't find the installation instructions. There is a README in the git/ subdirectory but it is 5 year old. So it mustn't be it.

I'm sorry, I'm probably missing something obvious,.

pedrobaeza:
Look in https://odoo-community.org/blog/the-oca-blog-1/post/black-isort-pre-commit-97

@sbidoul for me it's OK to remove that directory or put a big deprecation warning about that.

dreispt:
Got it, thanks Pedro. I was expecting to have it on this repo.

I agree we should remove the deprecated pre-commit.
We should also add the blog link to the main README.
Want help on that?

sbidoul:
> Want help on that?

Sure, it's most welcome :)

dreispt:
👉  #657

PR #672, opened 2020-06-30
Install MQT dependencies in PIP mode too

Yajo:
This should re-enable tours everywhere where PIP mode is being used and tours are being skipped.

See https://github.com/OCA/social/pull/560#issuecomment-651704674 and https://github.com/OCA/social/pull/555#issuecomment-651687647.

@sbidoul @Tecnativa TT24523

### DIFF
--- a/travis/travis_install_nightly
+++ b/travis/travis_install_nightly
@@ -117,6 +117,8 @@ if [[ "${MQT_DEP}" == "OCA" ]] ; then
     clone_oca_dependencies
 else
     pip install --upgrade pip setuptools wheel
+    # Needed for tour tests to not be skipped
+    pip install --upgrade websocket-client
 
     echo "Installing Odoo"
     pip install -r ${ODOO_PATH}/requirements.txt -e ${ODOO_PATH}


Yajo:
Expect many branches to go red after this is merged, but... :shrug:

sbidoul:
What is the missing library? The proper way to fix it is by adding it to `test-requirements.txt` in each repo where it's needed.

pedrobaeza:
@sbidoul see links for more info, but that's not an specific requirement for the module, but an Odoo one, so it should be part of the "MQT pack", being in PIP mode or classic one.

sbidoul:
I think it's better to put `websocket-client` in `test-requirements.txt` in repos where it's needed, to be more precise and
- not depend on a global list that we can't adapt depending on the Odoo version
- make it easier for people to discover requirements needed to run test locally
- in general reduce the amount of magic required to run tests

pedrobaeza:
I don't think so, because it's a requirement for website tours. That's why there's a WEBSITE_REPO environment variable for identifying all this stuff. Skipping tours silently is very risky for forcing contributors to add that obscure thing for getting it work. The PIP mode is supposed to work as MQT did, and here is not true, so please restore it or remove this mode.

Yajo:
`WEBSITE_REPO` is only used to install things such as ruby sass and less: https://github.com/OCA/maintainer-quality-tools/blob/58da1d432e51a98db700f33469ff1bfdd6a7913f/travis/travis_install_nightly#L34-L46

It should not be needed anymore in v12+, where odoo uses libsass. Only in the weird corner case of any theme that uses less or ruby sass.

Other than that, having to specify that a repo supports running tours is the best way to have false positives. Tours should be supported everywhere, and by default.

OTOH, how is it possible that MQT was working in PIP mode if MQT's requirements were not installed?

Finally, `websocket-client` is a soft requirement by odoo itself to run frontend tests, so nothing related to the specific repo or test itself. Based on that, I agree with @pedrobaeza and it should be installed always and by default.

sbidoul:
> how is it possible that MQT was working in PIP mode if MQT's requirements were not installed

Odoo requirements are installed. I explicitly did not install MQT's requirement.txt in PIP mode because I wanted to understand why they are needed and make the test dependencies more explicit on a per-repo basis, for the reasons I explained above.

> websocket-client is a soft requirement by odoo itself to run frontend tests

Ok, put like that, it can make sense to install it globally.

I would be nicer if Odoo could have some test-requirement.txt in their repo to make such soft test requirements explicit for each series.

Yajo:
OK, then should we merge?

sbidoul:
I'd prefer to install websocket-client explicitly in PIP mode with a comment explaining why it is needed.
The rest of `requirements.txt` I still have no clear idea why it is necessary, so we can analyze case base case if/when problems arise and decide for each if we install it globally or locally in the repo.

pedrobaeza:
I have changed the PR to install that library in PIP mode. Please check and merge for solving the problem.

sbidoul:
Thanks!

PR #673, opened 2020-07-26
[REF] pylint_pr.cfg: Enable po checks

moylop260:
Enable pylint-odoo checks: duplicate-po-message-definition, po-msgstr-variables and po-syntax-error
In order to detect PO files errors early

More info about these checks:
 - https://github.com/OCA/pylint-odoo/pull/290

It is enabled as beta so won't affect build status


moylop260:
@pedrobaeza @sbidoul @dreispt 
Could you check it, please?

pedrobaeza:
We should enable them as well in pre-commit for 13.0

sbidoul:
> We should enable them as well in pre-commit for 13.0

Agreed, this should be done in https://github.com/OCA/oca-addons-repo-template

moylop260:
So, Can we merge it?

PR #676, opened 2020-09-23
IMP: apply autoflake to remove unsed imports

celm1990:
Before, unsed import must be removed manually by developer, but with this hook this task can be automated

is possible propagate this hook to every repo on OCA???

image

What do you think  @sbidoul @Yajo  @pedrobaeza

### DIFF
--- a/sample_files/pre-commit-13.0/.pre-commit-config.yaml
+++ b/sample_files/pre-commit-13.0/.pre-commit-config.yaml
@@ -13,6 +13,11 @@ exclude: |
 default_language_version:
   python: python3
 repos:
+  - repo: https://github.com/myint/autoflake
+    rev: master
+    hooks:
+      - id: autoflake
+        args: [ "-i", "--remove-all-unused-imports", "--ignore-init-module-imports" ]
   - repo: https://github.com/psf/black
     rev: 19.10b0
     hooks:


pedrobaeza:
This is interesting, but it can be applied only to 13.0/14.0 branches. There's no harm in doing it in 13.0, as linter already prevents to leave unused imports, so no red branches.

ivantodorovich (approved):

celm1990:
> This is interesting, but it can be applied only to 13.0/14.0 branches. There's no harm in doing it in 13.0, as linter already prevents to leave unused imports, so no red branches.

Ok, so i can create 1 PR on every repo(for branch 13.0 and 14.0 when this is release)???

pedrobaeza:
It's not needed. We have bots for that, but I'm wondering that current bots don't serve for the duality of configurations we will have on 13.0 and 14.0, and it should go more through https://github.com/OCA/oca-addons-repo-template

Yajo:
Yes, please open the corresponding PR on https://github.com/OCA/oca-addons-repo-template, which is going to be the future of config file maintenance here at OCA (well, I'm not sure if it's the present already). It'd be nice if you manage to do that before v14 is rolled out. CC @sbidoul

sbidoul:
I plan to work on a mechanism to auto-update the repo dot files from https://github.com/OCA/oca-addons-repo-template in the coming weeks. 

We probably want to deprecate the `sample_files` directory in this repo.

So I'm going to close this PR.

ivantodorovich:
> I plan to work on a mechanism to auto-update the repo dot files from https://github.com/OCA/oca-addons-repo-template in the coming weeks. 
> 
> 
> 
> We probably want to deprecate the `sample_files` directory in this repo.
> 
> 
> 
> So I'm going to close this PR.
> 
> 
> 
> 

In case you're interested, we sync dot files using GitHub actions like this: https://github.com/druidoo/druidoo-repo-sync

sbidoul:
@ivantodorovich thanks for the pointer, I'll have a look.

My plan was to use the `copier` update feature and integrate that in the `oca-github-bot`. And maybe do a PR if pre-commit fails.

PR #677, opened 2020-10-07
In pip install mode, support empty addons list

sbidoul:
Do not fail if there is no addon to test.


sbidoul:
Fast tracking trivial change

PR #678, opened 2020-10-14
Remove sample_files directory

sbidoul:
Replaced by OCA/oca-addons-repo-template

PR #685, opened 2021-01-15
[RFR] pre-commit: prevent pyupgrade to replace printf by str.format()

StefanRijnhart:
So as not to encourage developers to use str.format() as there are cases
where it can introduce security issues.

See https://github.com/OCA/pylint-odoo/issues/302

### DIFF
--- a/sample_files/pre-commit-13.0/.pre-commit-config.yaml
+++ b/sample_files/pre-commit-13.0/.pre-commit-config.yaml
@@ -90,6 +90,7 @@ repos:
     rev: v1.26.2
     hooks:
       - id: pyupgrade
+        args: [--keep-percent-format]
   - repo: https://github.com/pre-commit/mirrors-isort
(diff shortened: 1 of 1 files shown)


ivantodorovich (approved):

moylop260 (approved):
👍

pedrobaeza:
Aren't these sample files deprecated and this should be done on the copier template?

cc @Yajo

StefanRijnhart:
Ah yes I was wondering about whether this was the right place to update the files. Is this the right place then? https://github.com/OCA/oca-addons-repo-template/blob/master/.pre-commit-config.yaml.jinja#L116

pedrobaeza:
Yes, that's it, and #678 removes these samples.

StefanRijnhart:
Thanks. Superseded by https://github.com/OCA/oca-addons-repo-template/pull/49 then.

PR #691, opened 2021-04-12
Upgrade setuptools

sbidoul:
Using a fresh setuptools helps in some case, as it obey python-requires properly, and more and more packages have recent releases that require python >= 3 and must be avoided for python 2.


rousseldenis (approved):

sbidoul:
This fixes https://github.com/OCA/sale-reporting/pull/108. So merging.

PR #694, opened 2021-08-18
Use a personal access token to push to github

sbidoul:
@pedrobaeza @simahawk


sbidoul:
@pedrobaeza @simahawk the good news is that this syntax is correct. I tried with a new token.
The bad news is that the token that is set on all repos does not seem to work anymore...

Now, how do we change the token in all repos ? :cry:

sbidoul:
ah, there is an old PR: https://github.com/OCA/maintainer-tools/pull/349

PR #701, opened 2021-12-06
[NEW] Setup database template / unaccent on travis and runbot

mileo:
Hi,

We are trying to run OCA travis tests with the unaccent extension:

https://github.com/OCA/l10n-brazil/pull/1701/files#diff-6ac3f79fc25d95cd1e3d51da53a4b21b939437392578a35ae8cd6d5366ca5485

It was only possible when we added this new optional parameter that allows you to use a template database in the test setup.

### DIFF
--- a/travis/test_server.py
+++ b/travis/test_server.py
@@ -210,7 +210,7 @@ def cmd_strip_secret(cmd):
 
 
 def setup_server(db, odoo_unittest, tested_addons, server_path, script_name,
-                 addons_path, install_options, preinstall_modules=None,
+                 addons_path, install_options, setup_dbtemplate, preinstall_modules=None,
                  unbuffer=True, server_options=None):
     """
     Setup the base module before running the tests
@@ -222,6 +222,7 @@ def setup_server(db, odoo_unittest, tested_addons, server_path, script_name,
     :param script_name: name of the main server file
     :param addons_path: Addons path
     :param install_options: Install options (travis parameter)
+    :param setup_dbtemplate: use an alternative database template
     :param preinstall_modules: (list) Modules that should be preinstalled
     :param unbuffer: keeps output colors
     :param server_options: (list) Add these flags to the Odoo server init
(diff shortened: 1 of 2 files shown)


mileo:
Unfortunately the unaccent fix does not propagate to the runbot, would it be necessary to add these environment variables to the runbot?

https://runbot2.odoo-community.org/runbot/static/build/3520430-1701-228d66/logs/job_20_test_all.txt

CC: @moylop260 @gurneyalex 

Thanks in advance

gurneyalex:
runbot is dead. Please open a new issue on runboat if needed.

gurneyalex (commented):
ok with me

gurneyalex (approved):

rvalyi:
@mileo before we depend on Travis specific things while Travis is already dead, I approve this if you propose something equivalent for Runboat here https://github.com/OCA/oca-ci/tree/master/bin and if it gets approved.
Otherwise it's going to be a shot in the foot for the migration from v12...

mileo:
> @mileo before we depend on Travis specific things while Travis is already dead, I approve this if you propose something equivalent for Runboat here https://github.com/OCA/oca-ci/tree/master/bin and if it gets approved. Otherwise it's going to be a shot in the foot for the migration from v12... Alternatively the test can probably be written with a mock to avoiding tweaking the Porstgres CI.

We will work on that ASAP.

mileo:
> @mileo before we depend on Travis specific things while Travis is already dead, I approve this if you propose something equivalent for Runboat here https://github.com/OCA/oca-ci/tree/master/bin and if it gets approved. Otherwise it's going to be a shot in the foot for the migration from v12... Alternatively the test can probably be written with a mock to avoiding tweaking the Porstgres CI.

I have verified that many OCA projects still using travis (13/14) and i don't think i's a problem to merge it.

And from what I analyzed in the runboat/oca-ci code it is simpler and less parameterizable than maintainer-quality-tools keeping the same db, which apparently won't cause the same problem, but it's necessary to test to have 100% of assurance.

PR #706, opened 2022-05-06
[REF] run_pylint: Compatibility with pylint>2.12

moylop260:
The following commit:
 - https://github.com/OCA/maintainer-quality-tools/commit/365183598bc86889de0349aa23d015cc5f356577

It is compatible with new way to get result of messages based on:
 - https://github.com/PyCQA/pylint/commit/16c09cb#diff-3a681c16a79ae9818f939e08df8b9bb92ef9ce6d5fc65c360d7c9b526c3c3410R507

However, new corner cases were raised

e.g. https://github.com/OCA/sale-workflow/pull/2004#issuecomment-1118506829

It is because the test_pylint script consider returns early using a dict

So, we need to transform new way object to old way dict in order to be compatible with this old script
with minimal changes since that new pylint checks are available from pre-commit

### DIFF
--- a/travis/run_pylint.py
+++ b/travis/run_pylint.py
@@ -168,8 +168,8 @@ def pylint_run(is_pr, version, dir):
 
     real_errors = main(cmd, standalone_mode=False)
     res = dict(
-        (key, value) for key, value in (real_errors.by_msg
-            or {}).items() if key not in beta_msgs)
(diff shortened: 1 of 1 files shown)


moylop260:
Tested from https://github.com/OCA/sale-workflow/pull/2013

PR #707, opened 2022-05-26
[UPD] makepot: add 14.0 and 15.0 support

MiquelRForgeFlow:
Please! :)

### DIFF
--- a/travis/test_server.py
+++ b/travis/test_server.py
@@ -476,7 +476,8 @@ def main(argv=None):
         and os.environ.get('TRAVIS_REPO_SLUG', '').startswith('OCA/')
         and (
             os.environ.get('TRAVIS_BRANCH')
-            in ('8.0', '9.0', '10.0', '11.0', '12.0', '13.0')
+            in ('8.0', '9.0', '10.0', '11.0', '12.0', '13.0', '14.0', '15.0',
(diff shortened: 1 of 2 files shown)


pedrobaeza (approved):
Ouch, this is very heavy that we are without POT generation. I didn't know this is by branches. Please add 16.0, 17.0 and 18.0 in advance for avoiding this problem in the future.

AaronHForgeFlow:
wtf

MiquelRForgeFlow:
@pedrobaeza checks are green

pedrobaeza:
Not sure if this needs to be deployed somewhere. Maybe with GH actions is not needed? cc @sbidoul

sbidoul:
There is `or "ocabot-merge" in os.environ.get('TRAVIS_BRANCH', '')`

So .pot are generated when merging with the bot, right ?

pedrobaeza:
I can't really say... @MiquelRForgeFlow you have got to this due to something missing that traces back to here?

sbidoul:
In think the only "problem" is that .pot files are not regenerated nightly on the 14.0 and 15.0 branches.

When merging with the bot, they are generated on merge both with Travis and GitHub actions.

pedrobaeza:
And with bot, all POT files are regenerated, or only of the affected modules?

